PR #87, opened 2026-04-25
docs: add getting started guide

shivchander:
## Summary
- Adds `docs/getting-started.md` covering how to create a new project from scratch using the Factory
- Documents all four input methods: raw prompt, idea file, GitHub URL, existing directory
- Covers interactive vs headless mode, continuous loop, focus areas, and `factory.md` basics
- Adds "Quick Start" as the first nav item under "Getting Started" in mkdocs

## Test plan
- [x] All 873 tests pass
- [x] Ruff lint passes
- [x] mkdocs nav structure validates correctly
- [ ] Verify page renders on mkdocs serve

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -55,7 +55,7 @@ Clone and improve an existing repo:
 factory ceo https://github.com/user/repo
 ```
 
-The Factory clones the repo to a temp directory, discovers what it does, sets up evaluation dimensions, and starts improving it.
+The Factory clones the repo to a temporary directory, discovers what it does, sets up evaluation dimensions, and starts improving it. If you plan to keep the results, clone the repo yourself first and use the local directory path instead — temp directories don't survive reboots.
 
 ### From an existing directory
 
@@ -94,7 +94,7 @@ Headless mode is useful for scripting and automation.
 
 ## Continuous Improvement
 
-Run the Factory in a loop so it keeps improving your project unattended:
+`factory run` is equivalent to `factory ceo` but designed for unattended operation — run it in a loop so the Factory keeps improving your project:
 
 ```bash
 factory run ~/my-project --loop                    # every 30 min (default)
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -24,7 +24,7 @@ factory ceo "Build a CLI that converts CSV to JSON with streaming support"
 
 This will:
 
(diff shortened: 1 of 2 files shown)


akashgit (commented):
## Review: docs/getting-started.md

Nice contribution, Shiv. The guide is well-structured and covers the four input methods clearly. I verified all CLI flags, defaults, and cross-references against the codebase — everything checks out.

### Accuracy ✓

- `factory ceo <prompt>` → raw prompt path in `_resolve_input` — correct
- `FACTORY_PROJECTS_DIR` env var — exists, default `~/cursor-projects` — correct
- Slug truncation to 50 chars — matches `_slugify` — correct
- `--headless`, `--focus`, `--loop`, `--interval 1800` (30 min default), `--max-cycles` — all match argparse definitions
- `factory tmux`, `tmux-ls`, `tmux-stop` — all exist
- `factory.md` example uses `## Scope / ### Modifiable`, `## Guards`, `## Eval / ### Threshold` — all are recognized config sections parsed by `store.py`
- Cross-referenced pages (`setup.md`, `configuration.md`, `architecture.md`, `eval.md`, `self-improvement.md`) all exist in `docs/`
- Nav placement (Quick Start before Setup) makes sense

### Issues to address

1. **GitHub clone goes to a temp dir** — The guide says "The Factory clones the repo to a temp directory, discovers what it does..." but doesn't mention that `tempfile.mkdtemp()` is ephemeral. If the machine reboots, the project is gone. This should either (a) warn the user to copy/move the project afterward, or (b) suggest using `git clone` manually first and then pointing `factory ceo` at the local directory instead. As-is, a user could lose work.

2. **`factory run` vs `factory ceo` distinction is unclear** — The guide uses `factory run` in the "Continuous Improvement" section and `factory ceo` everywhere else, but doesn't explain the difference. A one-liner like "`factory run` is an alias for `factory ceo` with loop support" (or whatever the actual distinction is) would help.

### Nits

- The cycle description ("Detect → Research → Strategize → Build → Evaluate → Decide → Archive") is a good mental model but the actual CEO prompt uses slightly different terminology (Observe, Hypothesize, Implement, etc.). Not blocking, but worth aligning if there's a canonical list.

Overall: solid first docs contribution. Address the temp-dir footgun (#1) and this is good to merge.

PR #96, opened 2026-04-26
fix: correct factory ceo examples to use positional prompt

### LINKED ISSUE #91: README `factory ceo` example is out of date
The quickstart example in the README fails as written:

```bash
factory ceo --prompt "Build a REST API for bookmark management with tags and search"
```

Two issues with the current CLI:

1. **Missing required `path` positional argument.** Running the command above errors with:
```
   factory ceo: error: the following arguments are required: path
```

2. **`--prompt` expects a file path, not an inline string.** Passing a string causes the CLI to look for a file with that string as the name:
```
   Error: prompt file not found: /workspace/home/churwitz/cursor-projects/bookmark/Build a REST API for bookmark management with tags and search
```
   The `--help` text confirms: `Path to a prompt/spec file (absolute or relative to project). Loaded as the build spec into .factory/strategy/current.md`.

**Suggested fix** — update the README example to something like:

```bash
echo "Build a REST API for bookmark management with tags and search" > /tmp/spec.md
factory ceo ./bookmark-api --prompt /tmp/spec.md
```

akashgit:
## Summary
- Remove `--prompt` flag from all inline-string examples in README, CLAUDE.md, and docs/index.md
- The `--prompt` flag expects a **file path**, not an inline string — passing a raw string causes a "file not found" error
- The positional `path` argument already handles raw text prompts via `_resolve_input()` (case 4: slug + git init + persist spec)
- Also corrects the `--prompt` description in CLAUDE.md and updates stale test count in README

Fixes #91

## Test plan
- [x] `uv run pytest tests/ -v` — 939 passed
- [x] `uv run ruff check .` — clean
- [x] Verified `factory ceo "Build a test app"` works (positional prompt → `_resolve_input` case 4)
- [x] Verified no remaining `factory ceo --prompt "inline"` patterns: `grep -rn 'factory ceo --prompt "' --include='*.md'` returns 0 matches

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -95,7 +95,7 @@ Requires Claude Code installed and authenticated. The factory spawns `claude` su
 factory ceo /path/to/project                    # Launch CEO agent (single cycle)
 factory ceo /path/to/project --mode meta        # Improve + ACE playbook evolution
 factory ceo /path/to/project --focus "dashboard UI"  # Focus on a specific area
-factory ceo --prompt "Build a weather CLI"      # Build from a raw prompt
+factory ceo "Build a weather CLI"               # Build from a raw prompt
 factory run /path/to/project                    # Same as factory ceo
 factory run /path/to/project --loop --interval 1800  # Continuous heartbeat
 factory tmux /path/to/project --loop            # In detached tmux session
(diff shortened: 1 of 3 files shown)


akashgit (commented):
**Review: LGTM** ✓

Verified the core claim against the code:

- `--prompt` calls `_read_prompt_file()` (`cli.py:1176`) which treats the arg as a **file path** and exits with "file not found" if it doesn't exist — so `--prompt "Build a weather CLI"` would fail
- The positional `path` arg goes through `_resolve_input()` (`cli.py:1118`), case 4 handles raw text strings correctly (slugify → init repo → persist spec)

All doc changes are consistent and accurate. The `--prompt` description correction in CLAUDE.md ("loads a spec file" vs "builds from raw text") is also correct.

PR #149, opened 2026-04-30
fix: add backlog item verification gates to CEO prompt

akashgit:
## Summary

The CEO previously removed backlog items unconditionally — if a hypothesis was tagged `**Backlog item:**` and the experiment passed precheck, the item was deleted without checking whether the work actually solved the original ask. This allowed prerequisites to be counted as clearing operational backlog items (e.g., "wire code" counted as solving "run pipeline on real data").

Three verification gates:

1. **Strategy hard gate (Step 1r):** CEO reads original backlog item from `backlog.md`, compares against the hypothesis, and REDIRECTs if the hypothesis only partially addresses it. A prerequisite tagged as clearing an operational item gets caught here.

2. **Post-experiment (Step 2g keep block):** `backlog-remove` is now conditional. CEO reads the PR diff, E2E result, and execution artifacts to judge: YES (remove), NO (keep in backlog), or PARTIAL (update item text to reflect remaining work).

3. **End-of-cycle (Step 3b):** CEO reviews backlog after session summary to catch items that should have been cleared but weren't, and writes backlog status to the summary.

## Test plan

- [x] All 1144 tests pass
- [ ] Run factory improve cycle with a backlog item and verify CEO checks adequacy at strategy gate
- [ ] Verify a kept experiment that only partially addresses a backlog item leaves the item in the backlog


akashgit (commented):
## Review — LGTM after 1 fix

**Issue found and fixed:** The "Backlog item adequacy" check overlapped with "Operational item validation" for operational backlog items — both caught "hypothesis proposes code but backlog asks for execution." Tightened the adequacy check to explicitly note it covers ALL backlog items (the operational case is already handled by the check above), and added concrete non-operational mismatch examples (subset of features, missing UI, incomplete error handling) so the CEO knows what to look for beyond the execution case.

All three gates are clean:
- **Strategy gate:** CEO compares hypothesis against original backlog item text, REDIRECTs on partial coverage
- **Post-experiment:** backlog-remove is conditional on YES/NO/PARTIAL judgment with concrete actions for each
- **End-of-cycle:** CEO reviews remaining backlog after summary, flags items that should have been cleared

1144 tests pass. Ready to merge.

akashgit (commented):
## Review Round 2 — Fixed

Two issues found and fixed:

**1. Missing `backlog_cleared` in notes convention**
The `--notes` field and finalize examples didn't track whether the backlog item was verified as solved. Added `backlog_cleared=yes|no|partial|na` to the convention list and both keep/revert examples.

**2. Backlog verification logic was inside a bash comment block**
CEO judgment logic (read backlog, compare against PR, decide YES/NO/PARTIAL) was written as bash comments — invisible to the CEO since it treats comment blocks as shell instructions, not decision prompts. Moved the verification out of the bash block into structured prose with clear branches and inline bash for the commands.

All 1144 tests pass. LGTM — ready to merge.

PR #169, opened 2026-05-02
Fix Live Events panel expanding without bounds

### LINKED ISSUE #168: Dashboard: Live Events panel should be scrollable with fixed height
Image

## Problem

The Live Events panel in the Factory dashboard expands vertically without bounds as events accumulate. This pushes the Experiments table far below the fold, requiring the user to scroll the entire page to see experiment results.

In the current layout, the Projects sidebar, Live Events, and Experiments sections are stacked. When Live Events grows, the Experiments table becomes inaccessible without extensive scrolling.

## Expected behavior

The Live Events panel should have a fixed max-height with internal scrolling (`overflow-y: auto`), keeping the Experiments table visible without full-page scrolling. Ideally the layout would be:

- Projects sidebar: fixed height, scrollable if many projects
- Live Events: fixed max-height (e.g. 400px), scrolls internally, auto-scrolls to newest event
- Experiments table: always visible below Live Events without page scroll

## Environment

- Factory v0.2.0
- Browser: Chrome (macOS)
- Dashboard served via `factory dashboard --projects-dir /path/to/projects`

lukeinglis:
## Summary

- Add `min-height: 0` to `.panel` CSS class to fix unbounded vertical growth of the Live Events panel on desktop (grid `1fr` tracks)
- Add `max-height: 50vh` to `.events-panel` and `.experiments-panel` in the responsive breakpoint (`@media max-width: 800px`) to constrain panel height on narrow viewports where `auto` grid rows are used
- CSS grid items default to `min-height: auto`, which sizes them to fit content and prevents internal scrolling. The desktop fix overrides that default so `overflow-y: auto` on `.panel-body` engages correctly. The mobile fix uses explicit height constraints since `auto` tracks don't benefit from `min-height: 0`.
- Add regression test assertion for `min-height: 0` in `test_index_returns_html`

## Test plan

- [ ] Open dashboard with a project that generates many events
- [ ] Verify Live Events panel scrolls internally instead of pushing Experiments below the fold (desktop)
- [ ] Resize browser below 800px and verify Events and Experiments panels scroll internally (mobile)
- [ ] Verify Experiments table remains visible without page scroll
- [ ] Verify Projects sidebar scrolls correctly with many projects
- [ ] All 31 existing tests pass

Closes #168

### DIFF
--- a/factory/dashboard/static/index.html
+++ b/factory/dashboard/static/index.html
@@ -88,6 +88,7 @@
     border-bottom: 1px solid var(--border);
     display: flex;
     flex-direction: column;
+    min-height: 0;
   }
 
   .panel-header {
--- a/factory/dashboard/static/index.html
+++ b/factory/dashboard/static/index.html
@@ -405,8 +405,8 @@
     .kpi-strip { flex-wrap: wrap; }
     .kpi-card { min-width: calc(50% - 12px); }
     .projects-panel { grid-row: 2; max-height: 200px; }
-    .events-panel { grid-column: 1; grid-row: 3; }
-    .experiments-panel { grid-column: 1; grid-row: 4; }
+    .events-panel { grid-column: 1; grid-row: 3; max-height: 50vh; }
+    .experiments-panel { grid-column: 1; grid-row: 4; max-height: 50vh; }
   }
 </style>
 </head>
--- a/tests/test_dashboard.py
+++ b/tests/test_dashboard.py
@@ -101,6 +101,7 @@ class TestDashboardAPI:
         assert resp.status_code == 200
         assert "text/html" in resp.headers["content-type"]
         assert "Factory" in resp.text
+        assert "min-height: 0" in resp.text
 
     def test_list_projects(self, client: TestClient):
         resp = client.get("/api/projects")


akashgit (changes requested):
## Code Review

### Finding 1: Responsive breakpoint has the same unbounded expansion — BLOCKER

The fix adds `min-height: 0` to `.panel`, which works for the desktop grid layout (which uses `1fr` tracks). But the `@media (max-width: 800px)` query uses `grid-template-rows: auto auto auto auto` — with `auto` tracks, `min-height: 0` has no effect and the events panel still expands without bounds on narrow viewports.

The PR claims "All three panels (Projects, Events, Experiments) now stay within their grid tracks and scroll internally" — this is only true on desktop. The responsive layout needs explicit height constraints (e.g., `max-height` on `.events-panel` and `.experiments-panel`, or switching to `1fr` rows with a viewport-height constraint).

### Finding 2: No test coverage for the CSS fix — ISSUE

The existing test suite (`tests/test_dashboard.py`) has 31 tests but none verify rendered CSS. A basic assertion in the existing `test_index_returns_html` would prevent regressions:

```python
def test_index_returns_html(self, client: TestClient):
    resp = client.get("/")
    assert resp.status_code == 200
    assert "text/html" in resp.headers["content-type"]
    assert "Factory" in resp.text
    # Verify panel overflow fix is present
    assert "min-height: 0" in resp.text
```

### Finding 3: PR description overstates scope — NIT

The description should clarify the fix is desktop-only, or the responsive layout fix should be included in scope.

---

**Verdict: Do not merge as-is.** The core one-line fix is correct for desktop, but the responsive breakpoint is broken (same problem), and the PR's own claims aren't met. Please fix the responsive layout, add a regression test, and update the description.

lukeinglis:
Thanks for the review. All three findings addressed in 7b1bd7c:

**Finding 1 (responsive breakpoint):** Added `max-height: 50vh` to `.events-panel` and `.experiments-panel` in the `@media (max-width: 800px)` block. This follows the same pattern already used for `.projects-panel` (`max-height: 200px`) at that breakpoint. Using `50vh` keeps each panel to at most half the viewport, so both remain visible and scroll internally on narrow screens.

**Finding 2 (test coverage):** Added `assert "min-height: 0" in resp.text` to `test_index_returns_html`. All 31 tests pass.

**Finding 3 (PR description):** Updated to clarify both desktop and mobile fixes, and added a mobile resize step to the test plan.

akashgit (approved):
## Re-review after fixes

The second commit (`7b1bd7c5`) addresses both findings from the initial review:

1. **Responsive breakpoint fixed** — `max-height: 50vh` on `.events-panel` and `.experiments-panel` constrains them on narrow viewports. Combined with `.panel-body`'s `overflow-y: auto` and `flex: 1`, content scrolls internally. The `auto` grid rows size to content capped by `max-height`, which is the correct approach for a single-column mobile layout.

2. **Regression test added** — `assert "min-height: 0" in resp.text` in `test_index_returns_html`.

CSS chain is correct:
- **Desktop**: `1fr` rows + `min-height: 0` on `.panel` → panels shrink within fractional tracks, `.panel-body` scrolls
- **Mobile**: `auto` rows + `max-height: 50vh` on event/experiment panels → content capped, `.panel-body` scrolls

All 31 dashboard tests pass. LGTM — approved.

PR #195, opened 2026-05-07
feat: finalize gate + hard constraints

RobotSail:
## Summary

Code-level enforcement of precheck inside `cmd_finalize`. When the CEO calls `factory finalize --verdict keep`, precheck now runs automatically. If it fails, the verdict is overridden to revert. The CEO cannot bypass this.

### Problem

Previously, calling `factory precheck` was prompt-instructed only. The CEO could (and did) skip it, keeping experiments without quality validation. In our AdaptFM competition run, the CEO optimized to 9.7x speedup across 40 cycles while IFEval was failing — meaning the submission scored ZERO.

### Solution

The finalize gate runs precheck inside `cmd_finalize` — the single chokepoint for all verdicts. No timing issues, no "did the CEO run precheck" uncertainty. It just runs.

Also adds `## Hard Constraints` in factory.md — user-defined shell commands enforced via precheck. Each must exit 0 or the experiment is reverted.

## Files

| File | Change |
|------|--------|
| `factory/models.py` | HardConstraint model + hard_constraints field on FactoryConfig |
| `factory/store.py` | _parse_hard_constraints() for factory.md |
| `factory/precheck.py` | check_hard_constraints() wired into run_precheck() |
| `factory/cli.py` | Finalize gate in cmd_finalize + verdict.overridden event |
| `tests/test_hard_constraints.py` | 20 tests — model, parser, precheck, finalize gate |

## Test plan

- [x] 20 tests pass
- [x] ruff clean

### DIFF
--- a/factory/cli.py
+++ b/factory/cli.py
@@ -263,8 +263,8 @@ def cmd_finalize(args: argparse.Namespace) -> int:
         config_path = project_path / ".factory" / "config.json"
         if config_path.exists():
             config = FactoryConfig(**json.loads(config_path.read_text()))
-            history = _run(store.load_history()) if hasattr(store, "load_history") else []
-            history_dicts = [r.model_dump() if hasattr(r, "model_dump") else r for r in history]
+            history = _run(store.load_history())
+            history_dicts = [r.model_dump() for r in history]
 
             precheck_result = run_precheck(
                 score_before=score_before,
@@ -274,7 +274,7 @@ def cmd_finalize(args: argparse.Namespace) -> int:
                 history=history_dicts,
                 project_path=project_path,
                 smoke_test_command=config.smoke_test,
-                hard_constraints=config.hard_constraints or None,
+                hard_constraints=config.hard_constraints,
             )
 
             if not precheck_result.passed:
--- a/factory/cli.py
+++ b/factory/cli.py
@@ -248,13 +248,47 @@ def cmd_begin(args: argparse.Namespace) -> int:
 
 
 def cmd_finalize(args: argparse.Namespace) -> int:
+    from factory.precheck import run_precheck
     from factory.store import ExperimentStore
-    from factory.models import ExperimentRecord
+    from factory.models import ExperimentRecord, FactoryConfig
 
     project_path = Path(args.path)
     store = ExperimentStore(project_path)
     score_before = getattr(args, "score_before", None)
     score_after = getattr(args, "score_after", None)
+    verdict = args.verdict
+    notes = args.notes or ""
+
+    if verdict == "keep":
+        config_path = project_path / ".factory" / "config.json"
+        if config_path.exists():
+            config = FactoryConfig(**json.loads(config_path.read_text()))
+            history = _run(store.load_history()) if hasattr(store, "load_history") else []
+            history_dicts = [r.model_dump() if hasattr(r, "model_dump") else r for r in history]
+
+            precheck_result = run_precheck(
+                score_before=score_before,
+                score_after=score_after,
+                threshold=config.eval_threshold,
+                hypothesis=args.hypothesis or "",
+                history=history_dicts,
+                project_path=project_path,
+                smoke_test_command=config.smoke_test,
+                hard_constraints=config.hard_constraints or None,
+            )
+
+            if not precheck_result.passed:
+                verdict = "revert"
+                failure_detail = "; ".join(precheck_result.blocking_failures)
+                notes = f"[OVERRIDDEN by finalize gate] precheck failed: {failure_detail}. {notes}"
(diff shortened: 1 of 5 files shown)


akashgit (approved):
Solid feature — the finalize gate is the right chokepoint for enforcing hard constraints. The AdaptFM incident (9.7x speedup, IFEval failing, zero score) makes the motivation clear. Code is clean, tests are thorough (20 tests covering model, parser, precheck integration, and the finalize gate itself).

**Issues fixed in 8d1526b:**

1. **Dead `hasattr` guards in `cmd_finalize`** — `hasattr(store, "load_history")` always evaluates to `True` since `load_history` is a method on `ExperimentStore`. The `else []` branch was unreachable. Same for `hasattr(r, "model_dump")` — `r` is always an `ExperimentRecord`. Removed both guards.

2. **Needless `or None` conversion** — `config.hard_constraints or None` converts `[]` to `None`, but `run_precheck` already handles both via `if hard_constraints:`. Passing `[]` directly is cleaner and avoids the implicit falsy-list-to-None conversion.

3. **Missing structured logging** — `check_hard_constraints` was the only check function in `precheck.py` without structured logging. Added `hard_constraints_start` and `hard_constraint_result` log events for observability parity with `check_smoke_test` et al.

akashgit (changes requested):
**Follow-up:** My earlier approval mentioned fixes in commit 8d1526b, but that commit landed on a local branch, not on this PR (fork-based). The issues below still need to be fixed.

### Issues to fix

1. **`store.py` — unsorted import** (line 14): `HardConstraint` is placed before `CompositeScore`, breaking the alphabetical order of the existing import block. Move it between `FactoryConfig` and `HypothesisBudget`.

2. **`cli.py:266` — dead `hasattr` guard**: `hasattr(store, "load_history")` always evaluates to `True` since `load_history` is defined on `ExperimentStore`. The `else []` branch is unreachable. Same for `hasattr(r, "model_dump")` on line 267 — `r` is always an `ExperimentRecord`. Remove both guards:
   ```python
   history = _run(store.load_history())
   history_dicts = [r.model_dump() for r in history]
   ```

3. **`cli.py:277` — needless `or None` conversion**: `config.hard_constraints or None` converts `[]` to `None`, but `run_precheck` already skips via `if hard_constraints:`. Pass `config.hard_constraints` directly.

4. **`precheck.py` — missing structured logging**: `check_hard_constraints` is the only check function without structured logging. Add log events for observability parity with `check_smoke_test` etc.

akashgit (approved):
## Review: feat: finalize gate + hard constraints

**Verdict: APPROVE**

### What this does

Embeds precheck directly into `cmd_finalize` — the single chokepoint for all experiment verdicts. When the CEO calls `factory finalize --verdict keep`, precheck runs automatically. If it fails, the verdict is overridden to `revert`. The CEO cannot bypass this. Also adds `HardConstraint` — user-defined shell commands in factory.md that must exit 0 for a keep to stand.

This is the right fix for the AdaptFM-style failure where the CEO optimized for speedup while IFEval was failing silently.

### Code quality

- **Tests**: 20 tests covering all four layers (model, parser, precheck integration, finalize gate e2e). Full suite (1450 tests) passes with no regressions.
- **Lint**: ruff clean on all changed files.
- **Type check**: mypy clean on changed files (pre-existing `mcp_server.py` errors only).
- **Architecture**: Follows existing patterns — lazy imports in cli.py, `ConfigDict(strict=True, extra="forbid")` on models, structlog usage, `shell=True` for user-defined commands (same as `smoke_test`).

### Design notes

1. **Chokepoint placement is correct.** `cmd_finalize` is the only path to record a verdict — putting the gate here means it's truly non-bypassable.

2. **Finalize gate runs a subset of precheck** — score direction, anti-pattern, smoke test, and hard constraints. Scope/surface/leakage checks are skipped because `baseline_sha` isn't available in `cmd_finalize`. This is a reasonable trade-off: the gate catches the most critical failures; the CEO should still run full `factory precheck` for defense-in-depth. Worth documenting this intent in a code comment on the `run_precheck` call in `cmd_finalize` so future readers know it's deliberate (not a missing parameter).

3. **Parser handles colons in check commands correctly** — `split(":", 1)` means `check: curl -sf http://localhost:8080/ping` parses cleanly.

4. **`config_path.exists()` silently skips the gate** when no config.json exists. This is tested and intentional, but means a project without discovery/config gets no finalize gate protection. Acceptable for the MVP since unconfigured projects haven't set up hard constraints anyway.

### One suggestion

Add a brief inline comment in `cmd_finalize` noting that the precheck call intentionally omits `baseline_sha`/`fixed_surfaces`/`allowed_scope` because they require git state not available at finalize time. Prevents a future contributor from treating it as a bug.

Clean, well-tested, solves a real problem. Ship it.

PR #197, opened 2026-05-07
fix: force-add agents/ in plugin workflow to bypass .gitignore

xukai92:
## Summary
- The `agents/` directory is in `.gitignore` (generated output), so `git add agents/` silently skips it
- Use `git add -f agents/` to force-add despite the gitignore rule

## Test plan
- [ ] Verify the plugin workflow succeeds on next push to main

### DIFF
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -32,7 +32,7 @@ jobs:
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git checkout -B plugin
-          git add agents/
+          git add -f agents/
           git diff --cached --quiet && echo "No changes" && exit 0
           git commit -m "build: generate plugin agent files from $(git rev-parse --short main)"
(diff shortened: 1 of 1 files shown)


akashgit (approved):
**LGTM — approved.**

Verified against CI logs: all 3 recent plugin workflow runs (`25520138568`, `25519078348`, `25518942161`) failed at the same point:

```
The following paths are ignored by one of your .gitignore files:
agents
hint: Use -f if you really want to add them.
##[error]Process completed with exit code 1.
```

`git add -f` is the correct fix for intentionally committing generated files that are gitignored in the working tree.

**One minor note on the PR description:** The root `.gitignore` has `.claude/agents/` (line 13), not bare `agents/`. That pattern shouldn't match top-level `agents/` by standard gitignore rules — the actual match may come from a CI-side global gitignore or a git version behavior difference (CI runs git 2.53.0). Doesn't affect the fix since `-f` bypasses all ignore rules regardless of source, but worth noting for posterity.

PR #198, opened 2026-05-07
feat: add pipeline skills for dynamic multi-agent orchestration

xukai92:
## Summary

- Add `/factory:pipeline` skill — designs a DAG of agent steps, executes them via `factory agent <role>` CLI calls with sequential + parallel batches
- Add `/factory:pipeline-subagents` skill — same pipeline design but spawns `factory-*` subagents via Claude Code's Agent tool (native parallel + background support)
- Both follow the same two-phase pattern: Design (analyze goal → produce step table with dependencies and gates) → Execute (topological order with PROCEED/REDIRECT/ABORT between steps)
- Skills only — no new agent roles, no Python changes, just skill files + tests

## Test plan

- [x] `uv run pytest tests/test_pipeline_prompt.py -v` — 22 structural tests for both skills
- [x] `uv run pytest -x -q` — full suite passes (1505 tests)
- [x] `uv run ruff check .` — lint clean
- [ ] Manual: invoke `/factory:pipeline "research and fix the auth bug"` in a Claude Code session

### DIFF
--- a/skills/pipeline-subagents/SKILL.md
+++ b/skills/pipeline-subagents/SKILL.md
@@ -1,6 +1,7 @@
 ---
 name: pipeline-subagents
-description: "Design and execute a custom multi-agent pipeline using Claude Code subagents directly. Spawns factory-researcher, factory-builder, etc. via the Agent tool with native parallel and background execution. Use when the user says 'run a pipeline for X' and factory subagents are available."
+description: "Design and execute a custom multi-agent pipeline using Claude Code subagents directly. Spawns researcher, builder, etc. via the Agent tool with native parallel and background execution. Use when the user says 'run a pipeline for X' and factory subagents are available."
+disable-model-invocation: true
 argument-hint: "<goal>"
 ---
 
@@ -10,44 +11,51 @@ You design and execute custom multi-agent pipelines using Claude Code's native A
 
 The user wants: **$ARGUMENTS**
 
+## Setup
+
+```bash
+mkdir -p .factory/pipeline
+```
+
 ## Your Agents
 
-Spawn specialists using the **Agent tool**:
+Spawn specialists using the **Agent tool** with the plugin-namespaced subagent type `factory:<role>`:
 
 ```
 Agent({
   description: "<short description>",
   prompt: "<detailed task>",
-  subagent_type: "factory-<role>"
+  subagent_type: "factory:<role>"
 })
 ```
 
 | Subagent Type | Purpose |
 |---------------|---------|
-| factory-researcher | Web research, codebase analysis, domain studies |
-| factory-strategist | Generate prioritized hypotheses from observations |
-| factory-builder | Implement code changes on a feature branch, open PRs |
-| factory-reviewer | Review PRs, guard checks, keep/revert verdicts |
-| factory-evaluator | Run evals, compare before/after scores |
-| factory-archivist | Record findings to `.factory/archive/` |
-| factory-distiller | Refine vague ideas into buildable specs |
-| factory-failure_analyst | Classify experiment failures by root cause |
+| factory:researcher | Web research, codebase analysis, domain studies |
+| factory:strategist | Generate prioritized hypotheses from observations |
+| factory:builder | Implement code changes on a feature branch, open PRs |
+| factory:reviewer | Review PRs, guard checks, keep/revert verdicts |
+| factory:evaluator | Run evals, compare before/after scores |
+| factory:archivist | Record findings to `.factory/archive/` |
+| factory:distiller | Refine vague ideas into buildable specs |
 
 ### Parallel Execution
 
 Issue multiple Agent tool calls in the **same message** — they run concurrently:
 
 ```
-Agent({ subagent_type: "factory-researcher", prompt: "Research the auth bug..." })
-Agent({ subagent_type: "factory-evaluator", prompt: "Run baseline eval..." })
+Agent({ subagent_type: "factory:researcher", prompt: "Research the auth bug..." })
+Agent({ subagent_type: "factory:evaluator", prompt: "Run baseline eval..." })
 ```
+
+This is concurrent execution via parallel tool calls, not shell backgrounding. Each Agent call is still individually synchronous.
 
 ### Background Execution
 
 For non-blocking steps (e.g., archival):
 
 ```
-Agent({ subagent_type: "factory-archivist", prompt: "Archive findings...", run_in_background: true })
+Agent({ subagent_type: "factory:archivist", prompt: "Archive findings...", run_in_background: true })
 ```
 
 ## Phase 1: Design the Pipeline
@@ -64,12 +72,12 @@ Agent({ subagent_type: "factory-archivist", prompt: "Archive findings...", run_i
 
 ### Steps
 
-| Step | Role | Task Summary | Depends On | Timeout |
-|------|------|-------------|-----------|---------|
-| S1 | researcher | ... | - | 300 |
-| S2 | evaluator | ... | - | 300 |
-| S3 | strategist | ... | S1, S2 | 300 |
-| ... | ... | ... | ... | ... |
+| Step | Role | Task Summary | Depends On |
+|------|------|-------------|-----------|
+| S1 | researcher | ... | - |
+| S2 | evaluator | ... | - |
+| S3 | strategist | ... | S1, S2 |
+| ... | ... | ... | ... |
 
 ### Gate Rules
 - After S1: PROCEED if ...; REDIRECT if ...
@@ -79,7 +87,7 @@ Agent({ subagent_type: "factory-archivist", prompt: "Archive findings...", run_i
 ### Design Principles
 
 - **Minimize invocations** — only agents needed for this goal
-- **Maximize parallelism** — steps with shared dependencies and no mutual dependency → spawn in same message
+- **Maximize parallelism** — steps whose dependencies are all satisfied and that don't depend on each other can be spawned in the same message
 - **Mandatory archival** — always include at least one archivist step at the end
 - **Gate rules** — define PROCEED/REDIRECT/ABORT criteria for critical transitions
 
--- a/skills/pipeline-subagents/SKILL.md
+++ b/skills/pipeline-subagents/SKILL.md
@@ -39,6 +39,8 @@ Agent({
 | factory:archivist | Record findings to `.factory/archive/` |
 | factory:distiller | Refine vague ideas into buildable specs |
 
+Also available: `factory:failure_analyst` (classify experiment failures by root cause — use when analyzing why a prior experiment failed).
+
 ### Parallel Execution
 
 Issue multiple Agent tool calls in the **same message** — they run concurrently:
--- a/skills/pipeline-subagents/SKILL.md
+++ b/skills/pipeline-subagents/SKILL.md
@@ -1,0 +1,110 @@
+---
+name: pipeline-subagents
+description: "Design and execute a custom multi-agent pipeline using Claude Code subagents directly. Spawns factory-researcher, factory-builder, etc. via the Agent tool with native parallel and background execution. Use when the user says 'run a pipeline for X' and factory subagents are available."
+argument-hint: "<goal>"
+---
+
+# Pipeline (Subagents) — Dynamic Multi-Agent Orchestrator
+
+You design and execute custom multi-agent pipelines using Claude Code's native Agent tool to spawn factory subagents directly.
+
+The user wants: **$ARGUMENTS**
+
+## Your Agents
+
+Spawn specialists using the **Agent tool**:
+
+```
+Agent({
+  description: "<short description>",
+  prompt: "<detailed task>",
+  subagent_type: "factory-<role>"
+})
+```
+
+| Subagent Type | Purpose |
+|---------------|---------|
+| factory-researcher | Web research, codebase analysis, domain studies |
+| factory-strategist | Generate prioritized hypotheses from observations |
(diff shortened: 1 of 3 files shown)


colehurwitz (commented):
# Code Review

## Overview

Adds two new Claude Code skills for ad-hoc multi-agent orchestration:
- `skills/pipeline/SKILL.md` — drives specialist agents via the `factory agent <role>` CLI.
- `skills/pipeline-subagents/SKILL.md` — drives them via Claude Code's native `Agent` tool with `subagent_type: factory-<role>`.

Both follow a Design → Execute pattern (DAG with PROCEED/REDIRECT/ABORT gates) and end with a mandatory archival step. +313 lines, no production-code changes; only skill docs and 22 structural keyword tests.

## 🔴 Critical: `pipeline-subagents` references subagent types that don't exist

`factory/agents/plugin.py:generate_agent_content` writes generated agent files with bare role names — there is no `factory-` prefix:

```python
{"name": role, "description": meta.description, ...}
```

Verified by running it:
```
---
name: researcher
description: Deep research and discovery for software projects...
```

But `skills/pipeline-subagents/SKILL.md` instructs Claude to invoke `subagent_type: "factory-researcher"`, `"factory-builder"`, `"factory-failure_analyst"`, etc. **None of those subagent_type values exist** — the actual values are `researcher`, `builder`, `failure_analyst`, etc. The skill as written will fail on every Agent call.

Two ways to fix:
- **Preferred:** rename the skill's references to the real names (`subagent_type: "researcher"` etc.) — matches what the plugin already ships.
- Otherwise: change `generate_agent_content` to emit `name: factory-<role>` and update every doc/test that references the bare names. Much wider blast radius.

The rest of this review assumes the first option.

## 🟠 Tests would have caught this — but they encode the bug

`tests/test_pipeline_prompt.py:80-81` actively asserts the wrong names:

```python
def test_references_factory_subagent_types(self, subagents_skill):
    assert \"factory-researcher\" in subagents_skill
    assert \"factory-builder\" in subagents_skill
```

So the test green-lit the broken contract. Suggestions:
- Replace these assertions with a check that reads the names from `agents.yml` (or from `factory.agents.plugin.load_agent_config()`) and asserts each appears in the skill. That way the skill drifts only when the source of truth changes.
- Add an analogous test for the CLI-based skill that reads valid roles from the `factory agent` argparse choices and asserts the skill's role table matches.

## 🟠 Missing directory creation before write

Both skills tell Claude to write `.factory/pipeline/plan.md` and `.factory/pipeline/summary.md`, but neither says to `mkdir -p .factory/pipeline` first. `.factory/` may not yet exist on a fresh project, and even if it does, the `pipeline/` subdir will not. Add the mkdir step to Phase 1 / Phase 2 close-out.

## 🟡 Smaller correctness / consistency issues

- **`PROJECT_PATH` is referenced but never set.** `skills/pipeline/SKILL.md` says \"Set `PROJECT_PATH` to the current working directory\" as prose, then uses `\"\$PROJECT_PATH\"` in every example. Either drop the variable and use `\"\$(pwd)\"` (matches `skills/implement/SKILL.md:29`) or include an explicit `PROJECT_PATH=\"\$(pwd)\"` in the prerequisites block.
- **Default timeout mismatch.** Skill table lists 300s for most roles; `factory agent --help` shows the CLI default is 600s. Either align with the CLI default or pass `--timeout 300` explicitly in examples. Right now an inattentive reader will assume 300s is the effective default.
- **Missing `disable-model-invocation: true`.** The neighboring `implement` skill (`skills/implement/SKILL.md:4`) sets this; pipeline skills don't. Without it the model can autonomously invoke pipelines from descriptions like \"orchestrate X\" — possibly desired, but worth a deliberate decision rather than an oversight.
- **Tension in pipeline skill section \"Invocation Rules\":** \"All invocations MUST be synchronous — no `&`, no `run_in_background`\" sits next to \"Issue multiple `factory agent` commands as separate bash tool calls in the same message turn. They execute concurrently.\" Both are technically true (concurrent ≠ background), but it reads as contradictory. A sentence clarifying \"concurrent parallel tool calls are fine; per-command shell backgrounding is not\" would help.
- **\"Maximize parallelism\" wording.** \"steps with shared dependencies and no mutual dependency → spawn in same message\" — the \"shared dependencies\" phrasing is misleading; the real condition is \"all dependencies satisfied AND no mutual dependency.\" Minor copy edit.
- **Distiller / failure_analyst aren't in `agents.yml` playbooks.** The skill table lists them; they exist in `prompts/` but only researcher/strategist/builder/reviewer/evaluator/archivist/ceo are in `playbooks/`. Worth confirming `factory agent distiller` and `factory agent failure_analyst` actually run end-to-end before recommending them in the skill.

## 🟢 What's good

- Two skills sharing a Design / Execute structure read cleanly and are easy to skim.
- Mandatory archival is called out as a design principle in both — consistent with the CEO's archival contract in CLAUDE.md.
- No Python changes, so the blast radius is bounded to skill files.
- Topological execution + gate protocol mirrors the existing CEO behavior, which keeps mental models consistent.

## Suggested changes before merge

1. **(blocker)** Fix the subagent_type names in `pipeline-subagents` to match the real generated names (drop `factory-` prefix).
2. **(blocker)** Update `tests/test_pipeline_prompt.py:80-81` so it reads the truth from `agents.yml` / argparse choices instead of hard-coding the wrong strings.
3. Add a `mkdir -p .factory/pipeline` step in both skills.
4. Replace `\$PROJECT_PATH` with `\$(pwd)` (or set the var explicitly).
5. Reconcile the 300s vs. 600s timeout claim against the CLI default.
6. Decide whether to add `disable-model-invocation: true` for parity with `implement`.
7. Verify `distiller` and `failure_analyst` actually run via `factory agent` before keeping them in the role table.

xukai92:
Thanks for the thorough review @colehurwitz — all feedback addressed in 63dad4d:

**Blockers (fixed):**
1. ✅ Subagent types now use the plugin namespace convention `factory:<role>` (e.g., `factory:researcher`). The `factory-` prefix was wrong.
2. ✅ Tests now read valid role names from `load_agent_config()` (agents.yml) instead of hardcoding. Added explicit regression guard against `factory-` prefix.

**Other fixes:**
3. ✅ Added `mkdir -p .factory/pipeline` to both skills
4. ✅ Replaced `$PROJECT_PATH` with `$(pwd)` to match the `implement` skill pattern
5. ✅ Removed timeout column from step table — CLI default (600s) is sufficient; agents can override per-step if needed
6. ✅ Added `disable-model-invocation: true` to both skills for parity with `implement`
7. ✅ Clarified "concurrent parallel tool calls ≠ shell backgrounding" in both skills

**Re: distiller/failure_analyst:** Both have prompts and are valid `AgentRole` entries, so `factory agent distiller` works. They just don't have evolved playbooks yet (playbooks are optional). Dropped `failure_analyst` from the subagents skill table since it's rarely needed in ad-hoc pipelines; kept `distiller` since it's useful for ideation pipelines.

colehurwitz (commented):
# Follow-up review on commit `fix: address PR #198 review feedback`

## What was fixed

Walked through the diff against my previous review — all 7 items addressed:

| # | Previous concern | Fix |
|---|------------------|-----|
| 1 | `factory-<role>` subagent_type didn't exist | Changed to `factory:<role>` plugin-namespaced form |
| 2 | Tests encoded the bug (`assert "factory-researcher" in ...`) | Tests now read from `load_agent_config()` and explicitly assert `factory-researcher` / `factory-builder` are NOT present |
| 3 | Missing `.factory/pipeline/` dir creation | `mkdir -p .factory/pipeline` added to both skills |
| 4 | `$PROJECT_PATH` referenced but never set | Replaced with `$(pwd)` |
| 5 | 300s vs 600s timeout mismatch | Timeout column dropped from role table — clean fix |
| 6 | `disable-model-invocation: true` missing | Added to both skills |
| 7 | Parallel-vs-background contradiction | Explicit clarifier added: "This is concurrent execution via parallel tool calls, not shell backgrounding. Each Agent call is still individually synchronous." |

Also dropped `failure_analyst` from the role tables (good — it's niche) and tightened the "Maximize parallelism" wording.

## Verified

- **`factory:<role>` is the correct subagent_type form.** Per Claude Code's official subagents docs, plugin-supplied subagents are referenced as `<plugin-name>:<agent-name>` — same colon-namespaced convention as plugin skills (`/factory:implement`). Since `.claude-plugin/plugin.json` declares `"name": "factory"` and `agents/researcher.md` has `name: researcher`, the resolved type is `factory:researcher`. The skill now matches that.
- **Tests pass:** `uv run pytest tests/test_pipeline_prompt.py -v` → 26 passed in 0.06s. The new assertions (`test_subagent_types_use_plugin_namespace`, `test_no_bare_or_dash_prefixed_subagent_types`, `test_creates_pipeline_directory`, `test_clarifies_parallel_vs_background`) actively guard against regression to the prior broken state.

## Small remaining notes (non-blocking)

- The two skills now have slightly different archival semantics — `pipeline-subagents` allows `run_in_background: true` for the archivist step, while `pipeline` (CLI) explicitly says all `factory agent` calls are synchronous. That divergence is real (the CLI runner has no background mode) and intentional, but worth a line in the docstring some day so a reader picking between the two skills understands why one is more flexible.
- My earlier note about distiller "not being in playbooks" was a false alarm on my part — playbooks are optional auto-injection (`~/.factory/playbooks/<role>.md`), not a prerequisite. Distiller has a prompt file and is a valid `factory agent` choice, so its inclusion in the role table is fine.
- The structural tests are good for catching this class of bug, but they don't smoke-test that an Agent invocation with `factory:researcher` actually resolves end-to-end. That's a fair line to draw — wiring a real plugin/Agent integration test into pytest is a different scope. Just flagging if anyone wants the next layer.

## Verdict

LGTM — ready to merge. The original critical bug is fixed, the test suite now guards against it, and the smaller polish items are all in. Nice turnaround.

xukai92:
Thanks for the thorough re-review @colehurwitz — glad everything checks out.

Re: the non-blocking notes:

1. **Archival semantics divergence** — agreed, that's intentional (CLI runner has no background mode). Will add a note in a follow-up if both skills stick around long-term.
2. **E2E integration test** — agreed it's a different scope. The structural tests catch contract bugs like the `factory-` prefix issue; wiring up a real plugin resolution test would need a mock Claude Code environment.

Merging!

akashgit (commented):
## PR #198 Review: Pipeline Skills for Multi-Agent Orchestration

**+350 lines, -0** | 3 new files | 26 tests passing, lint clean

### Overview

Adds two new Claude Code skills for ad-hoc multi-agent DAG orchestration:
- `/factory:pipeline` — drives agents via `factory agent <role>` CLI calls
- `/factory:pipeline-subagents` — drives agents via Claude Code's native `Agent` tool with `factory:<role>` plugin namespace

Both follow a two-phase Design → Execute pattern with topological ordering, parallel batching, PROCEED/REDIRECT/ABORT gates, and mandatory archival. No Python code changes — skills only.

The second commit addresses all critical issues from the first review round (subagent naming, mkdir, tests using `load_agent_config()`).

### Issues to Fix

**1. `factory:distiller` may not work as a plugin subagent**

The subagents skill lists `factory:distiller` in the agent table, but `load_agent_config()` only includes roles that have a corresponding prompt file in `factory/agents/prompts/`. Verify `distiller.md` exists there — if not, Claude Code will fail when trying to spawn `factory:distiller`. The test only checks the 6 core roles and would miss this.

**2. `failure_analyst` omitted from subagents but silently available**

The PR description says `failure_analyst` was dropped from the subagents table. Fair — but it's still in `agents.yml` and still available as a plugin subagent. If someone types a goal that involves failure analysis, the skill won't guide Claude toward using it. Consider adding a note like "additional roles available: `factory:failure_analyst`" or documenting the omission rationale in the skill itself.

**3. `$(pwd)` in pipeline skill won't resolve inside a subagent context**

`skills/pipeline/SKILL.md:28` uses `--project "$(pwd)"` in the example command. If the pipeline skill is invoked from a subagent that's running in a worktree or different CWD, `$(pwd)` will resolve to the wrong path. The `implement` skill uses `"$(pwd)"` too, so this is consistent — but worth noting that both would break in nested subagent scenarios. Low risk for now.

**4. Tests are all string-matching — no behavioral coverage**

All 26 tests assert keyword presence in markdown. These guard against accidental deletions but won't catch semantic drift. Consider adding at least one test that parses the step table format (regex for `| S\d+ |`) to validate the structural contract, not just keywords.

### Suggestions (non-blocking)

**5. No max-steps or token budget guidance** — A vague goal could generate a 15-step pipeline. Consider adding a soft cap: "Aim for 3-7 steps."

**6. Gate rules section could be more prescriptive** — Give a concrete example condition rather than leaving it fully free-form.

### Verdict

The fix commit properly addresses all issues from the first review. Skills are well-structured and consistent with the existing `implement` skill. Tests now read from `load_agent_config()` instead of hardcoding names.

Items 1-2 should be verified before merge; the rest are suggestions for future improvement. Solid contribution overall.

xukai92:
Thanks for the follow-up @colehurwitz — addressed in 7517e42:

1. **`factory:distiller` verified** — `distiller.md` exists in `factory/agents/prompts/` and `load_agent_config()` includes it. Confirmed both `distiller` and `failure_analyst` load correctly.
2. **`failure_analyst` documented** — added "Also available: `factory:failure_analyst`" note to the subagents skill so the pipeline agent knows it exists for failure analysis goals.
3. **`$(pwd)` in nested subagents** — acknowledged as consistent with `implement` skill. No change for now.
4. **Structural regex test added** — `test_step_table_has_structural_format` validates `| S\d+ |` rows exist, catching semantic drift beyond keyword presence.
5. **Max-steps / prescriptive gates** — good suggestions, will address in a follow-up.

PR #285, opened 2026-05-19
feat: tmux persist mode for CLI runners

xukai92:
## Summary

- Add `--tmux-persist` flag that launches agents interactively in tmux windows instead of headless subprocesses
- Agents run via `script` for output capture, with `tmux wait-for` for async blocking — factory gets `(stdout, return_code)` as before, but the user can attach to the tmux session and continue chatting after the agent finishes
- One tmux session per project (`factory-persist-<name>`), one window per agent invocation
- Configurable via CLI flag, `FACTORY_TMUX_PERSIST` env var, or `~/.factory/config.toml` (five-tier precedence)
- Falls back to normal headless mode when tmux is unavailable
- Bob runner logs a warning (no session resume support)

## Test plan

- [x] 16 unit tests in `test_tmux_persist.py` covering `tmux_available`, `_strip_ansi`, `run_in_tmux` (session creation, window reuse, timeout, ANSI stripping, error handling), and `ClaudeRunner` delegation/fallback
- [x] Updated mock signatures in `test_agents.py` for `tmux_persist` kwarg
- [x] Full suite: 1739 passed, 0 failed
- [ ] Manual: `factory agent researcher --task "list files" --project /tmp/test --tmux-persist`

### DIFF
--- a/factory/agents/runner.py
+++ b/factory/agents/runner.py
@@ -137,6 +137,7 @@ async def invoke_agent(
     _track_failures: bool = True,
     session_name: str | None = None,
     use_profile: bool = False,
+    tmux_persist: bool = False,
 ) -> tuple[str, int]:
     """Invoke a Claude Code agent with the resolved prompt + task.
 
@@ -153,6 +154,7 @@ async def invoke_agent(
         session_name: Optional session name for /resume identification.
             If not provided, defaults to "factory: {project_name}/{role}".
         use_profile: If True, inject user profile into the agent prompt.
+        tmux_persist: If True, run the agent interactively in a tmux window.
 
     Returns (stdout, return_code).
 
@@ -182,6 +184,7 @@ async def invoke_agent(
             dangerously_skip_permissions=dangerously_skip_permissions,
             role=role,
             session_name=agent_session_name,
+            tmux_persist=tmux_persist,
         )
     except Exception as e:
         logger.error("%s agent failed: %s", role, e)
@@ -282,6 +285,7 @@ async def invoke_agents_parallel(
     dangerously_skip_permissions: bool = True,
     model: str | None = None,
     runner_name: str | None = None,
+    tmux_persist: bool = False,
 ) -> list[tuple[str, int]]:
     """Invoke multiple agents concurrently. Returns list of (output, return_code).
 
@@ -299,6 +303,7 @@ async def invoke_agents_parallel(
             model=model,
             runner_name=runner_name,
             _track_failures=False,  # Avoid race condition; track locally below
+            tmux_persist=tmux_persist,
         )
         for role, task in tasks
     ]
--- a/factory/ceo_completion.py
+++ b/factory/ceo_completion.py
@@ -387,6 +387,7 @@ async def run_ceo_with_completion_guard(
     max_respawns: int | None = None,
     session_name: str | None = None,
     use_profile: bool = False,
+    tmux_persist: bool = False,
 ) -> tuple[str, int]:
     """Spawn CEO; if it exits with planned work undone, re-spawn until done or cap hit.
 
@@ -403,6 +404,7 @@ async def run_ceo_with_completion_guard(
         max_respawns: Max re-spawns (default from env or 5).
         session_name: Optional session name for /resume identification.
         use_profile: If True, inject user profile into the CEO prompt.
+        tmux_persist: If True, run agents in tmux windows.
 
     Returns:
         (final_output, exit_code)
@@ -419,6 +421,7 @@ async def run_ceo_with_completion_guard(
             timeout=timeout, model=model, runner_name=runner_name,
             session_name=session_name,
             use_profile=use_profile,
+            tmux_persist=tmux_persist,
         )
 
     if max_respawns is None:
@@ -459,6 +462,7 @@ async def run_ceo_with_completion_guard(
             timeout=timeout, model=model, runner_name=runner_name,
             session_name=session_name,
             use_profile=use_profile,
+            tmux_persist=tmux_persist,
         )
         final_output = result
 
--- a/factory/cli.py
+++ b/factory/cli.py
@@ -2177,6 +2177,7 @@ def cmd_agent(args: argparse.Namespace) -> int:
     model = _resolve_model(args)
     runner = _resolve_runner(args)
     use_profile = getattr(args, "use_profile", False)
+    tmux_persist = _resolve_tmux_persist(args)
 
     result, code = _run(invoke_agent(
         role,
@@ -2187,6 +2188,7 @@ def cmd_agent(args: argparse.Namespace) -> int:
         model=model,
         runner_name=runner,
         use_profile=use_profile,
+        tmux_persist=tmux_persist,
     ))
     print(result)
     return code
@@ -2369,6 +2371,7 @@ def cmd_ceo(args: argparse.Namespace) -> int:
     model = _resolve_model(args)
     runner_name = _resolve_runner(args)
     use_profile = getattr(args, "use_profile", False)
+    tmux_persist = _resolve_tmux_persist(args)
     clean_pr_flag = getattr(args, "clean_pr", None)
 
     if mode == "research" and not research_ideation and not _has_research_target(project_path):
@@ -2472,6 +2475,7 @@ def cmd_ceo(args: argparse.Namespace) -> int:
                 timeout=7200.0,
                 session_name=session_name,
                 use_profile=use_profile,
+                tmux_persist=tmux_persist,
             ))
             print(result)
             if code == 0:
@@ -2484,6 +2488,7 @@ def cmd_ceo(args: argparse.Namespace) -> int:
                 min_growth=min_growth, max_new=max_new, branch=branch,
                 already_improved=mode in ("improve", "meta") or discover_only,
                 model=model, no_github=no_github, use_profile=use_profile,
+                tmux_persist=tmux_persist,
             )
         finally:
             remove_worktree(project_path, wt_path, wt_branch)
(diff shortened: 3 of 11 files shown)


xukai92:
@codex can you review this?

akashgit (changes requested):
## Code Review — PR #285: tmux persist mode

### Overview

Adds a `--tmux-persist` flag that launches agents interactively in tmux windows instead of headless subprocesses, using `script` for output capture and `tmux wait-for` for async blocking. Well-structured feature with good test coverage (16 tests). However, there are two bugs that will prevent this from working on macOS, plus a gap that will crash the codex runner.

---

### Critical

#### 1. `script -c` is Linux-only — will fail on macOS

`_tmux_persist.py:82` generates:

```bash
script -q -c <cmd> <logfile>
```

The `-c` flag is a `util-linux` extension. macOS's `script` does not support it:

```
$ script -c "echo test" /dev/null
script: illegal option -- c
```

macOS syntax is `script -q <logfile> <command ...>`. The wrapper script needs platform detection:

```python
import platform
if platform.system() == "Darwin":
    f"script -q {shlex.quote(str(logfile))} {claude_cmd}\n"
else:
    f"script -q -c {shlex.quote(claude_cmd)} {shlex.quote(str(logfile))}\n"
```

#### 2. `CodexRunner.headless()` missing `tmux_persist` parameter — TypeError at runtime

The `Runner` protocol (`protocol.py`) and `invoke_agent` (`runner.py:142`) now pass `tmux_persist=tmux_persist` to `runner.headless()`, but `CodexRunner.headless()` was not updated to accept it. When using `--runner codex`, this will crash:

```
TypeError: CodexRunner.headless() got an unexpected keyword argument 'tmux_persist'
```

Fix: Add `tmux_persist: bool = False` to `CodexRunner.headless()` signature (with a warning like the bob runner, or just ignore it).

---

### Major

#### 3. `factory run` not wired up

`--tmux-persist` is added to the `factory agent` and `factory ceo` parsers, but not `factory run`. The `_run_single_cycle` function (line 2874) calls `invoke_agent` without `tmux_persist`. Since `factory run` is the primary entry point for continuous improvement, this is a gap in feature coverage.

#### 4. `timeout` silently ignored when tmux_persist is active

`ClaudeRunner.headless()` receives `timeout` from the caller but does not forward it to `run_in_tmux()`:

```python
# claude.py — timeout is available but not passed
return await run_in_tmux(
    prompt, task, cwd, role, _find_project_path(cwd),
    model=model,
    dangerously_skip_permissions=dangerously_skip_permissions,
)
```

`run_in_tmux` defaults to 86400s (24h). The caller's timeout (e.g. 7200s from `cmd_ceo`) is silently dropped. Either forward `timeout` or document the override.

---

### Minor

#### 5. `_find_project_path` belongs in `_tmux_persist.py`

This function is added as a module-level utility in `claude.py` but is only used for the tmux persist path. It duplicates project-root resolution logic. Move it to `_tmux_persist.py` where it's consumed.

#### 6. Session name collision

```python
session = f"{_SESSION_PREFIX}{project_path.name}"
```

Two projects at different paths with the same directory name (e.g. `/a/myapp` and `/b/myapp`) would share a tmux session. Consider hashing the full path or using a longer suffix.

---

### What looks good

- Clean separation: `_tmux_persist.py` is a well-scoped private module
- The `tmux wait-for` signaling pattern is solid — non-polling async wait
- ANSI stripping for clean output capture
- Graceful fallback when tmux is unavailable
- Bob runner warns instead of crashing
- Test coverage is thorough — session creation, window reuse, timeout, fallback, delegation
- Five-tier config precedence reused correctly via `_resolve_tmux_persist`

xukai92:
## Review feedback addressed

Rebased on upstream/main (merged cleanly with the new `session_name` parameter) and addressed all review items:

### Critical
1. **macOS `script -c` compatibility** — Added platform detection: Darwin uses `script -q <logfile> <cmd>`, Linux uses `script -q -c <cmd> <logfile>`
2. **CodexRunner missing `tmux_persist`** — No `CodexRunner` exists in the codebase (only `ClaudeRunner` and `BobRunner`). If it's added later, it will need `tmux_persist` in its signature.

### Major
3. **`factory run` not wired** — Added `--tmux-persist` to the `run` subcommand parser and threaded it through `_run_single_cycle`, `_chain_modes`, and the heartbeat loop
4. **Timeout silently ignored** — Now forwarded from the caller to `run_in_tmux()`

### Minor
5. **`_find_project_path` location** — Moved to `_tmux_persist.py` as `find_project_path()`
6. **Session name collision** — Appended SHA1 hash of the full project path: `factory-persist-myapp-a3f2b1`

All 2134 tests pass, lint clean.

colehurwitz (commented):
## Review: tmux persist mode

Nice direction — the wrapper-script + `tmux wait-for` design is a clean way to bridge headless capture and an attachable session. I read the full diff. One **blocker**, one **logic bug that defeats the feature's stated purpose**, and a few smaller things.

### 🔴 Blocker — breaks the Codex runner unconditionally

`factory/agents/runner.py` now passes `tmux_persist=tmux_persist` to `runner.headless(...)` on **every** invocation (line 184–185 of the diff), but `CodexRunner.headless` (`factory/runners/codex.py:61`) was not updated to accept the kwarg and has no `**kwargs` catch-all. Result: every codex-backend agent call will raise `TypeError: headless() got an unexpected keyword argument 'tmux_persist'`. It's caught by `invoke_agent`'s broad `except Exception`, so it surfaces as `"Error: ...", 1` — every codex agent silently fails. This isn't tmux-only; it's any user with `FACTORY_RUNNER=codex` or `--runner codex`.

The test suite passed (1739✓) because no test exercises `invoke_agent` against the real codex runner. Worth adding one to prevent this class of regression.

**Fix:** add `tmux_persist: bool = False` to `CodexRunner.headless` (and log a warning like bob does, since codex also has no interactive session model).

### 🟡 Interactive timeout silently truncates the session

`_DEFAULT_TMUX_TIMEOUT = 86400.0` in `_tmux_persist.py` is dead code in practice. `ClaudeRunner.headless` always forwards its own `timeout` (default 600s, CEO uses 7200s) into `run_in_tmux(..., timeout=timeout)`. So a user who attaches and continues chatting — the PR's headline use case ("user can attach to the tmux session and continue chatting after the agent finishes") — gets their window killed after 10 minutes (regular agent) or 2 hours (CEO), mid-conversation, by the `asyncio.wait_for` → `kill-window` path.

Two reasonable options:
1. When `tmux_persist=True`, ignore the caller's `timeout` and use the 24h default (or `None` / very large value).
2. Plumb a separate `interactive_timeout` through, or document that users must pass `--timeout` themselves.

Either way the current behavior contradicts what the PR description promises.

### 🟡 ANSI stripping is CSI-only

`_strip_ansi`'s regex (`\x1b\[[0-9;]*[a-zA-Z]`) handles CSI sequences but not the OSC title sequences (`\x1b]0;...\x07`), two-char escapes (`\x1b7`/`\x1b8` save/restore cursor), or DEC private mode sequences (`\x1b[?25l`) that the Claude TUI emits constantly. Captured output is fed back to the CEO for verdicts and persisted in `.factory/reviews/<role>-latest.md`; leaving escape junk in it can confuse downstream consumers.

Consider a more complete pattern, e.g. `re.compile(r"\x1b(\[[0-?]*[ -/]*[@-~]|\][^\x07]*\x07|[78=>])")`, or pulling in a small dep like `strip-ansi` / using `ansi2txt`.

### Minor

- **`_ensure_session` is a misnomer** — it `has-session`-checks, it doesn't ensure/create. Rename to `_session_exists` for clarity.
- **`find_project_path` duplication** — `bob.py` already has `_find_project_path` doing the same `.factory/` walk-up. Worth lifting into a shared util (`factory/runners/_paths.py` or similar) rather than maintaining two copies.
- **tmpdir leak on hard kill** — if the process is SIGKILL'd between `mkdtemp` and the `finally`, `/tmp/factory-tmux-*` accumulates. Not critical, but `tempfile.TemporaryDirectory` as a context manager would handle this; or have the wrapper script `rm -rf` its own dir last.
- **Concurrency + interactive is awkward** — `invoke_agents_parallel` now forwards `tmux_persist=True` to N agents, creating N windows in one session that all block on independent `wait-for` signals. Functionally fine, but a user attaching sees several parallel interactive prompts. Worth at least a docstring note, or auto-disable persist in the parallel path.
- **Protocol drift** — `Runner.headless` in `protocol.py` grew `tmux_persist`, so structurally `CodexRunner` no longer satisfies the protocol. mypy would catch this if any call site is `Runner`-typed; `get_runner`'s return type is worth checking.

### Test coverage

The 16 new tests are well-scoped — session-vs-window, timeout, ANSI stripping, fallback. Gaps that map to the issues above:
- No test for codex runner + invoke_agent (would have caught the blocker).
- No test pinning the 24h default actually being used in the interactive path.
- No test for OSC/non-CSI escape stripping.

### Summary

Solid feature with a clean implementation pattern, but as-is it (1) breaks the codex backend and (2) doesn't deliver the persistent-session UX promised in the description. Both are small fixes. Happy to re-review once the codex signature is updated and the timeout question is resolved.

xukai92:
## colehurwitz review addressed

Thanks for the thorough review! All items fixed:

### Blocker
- **CodexRunner crash** — Added `tmux_persist: bool = False` to `CodexRunner.headless()` with a warning log (same pattern as bob). Good catch — this was introduced by the Codex plugin merge after our initial implementation.

### Logic bug
- **Timeout truncates interactive session** — `ClaudeRunner.headless()` no longer forwards the caller's timeout to `run_in_tmux()`. The 24h default in `_tmux_persist.py` is now what's actually used for interactive sessions.

### ANSI stripping
- **CSI-only regex** — Replaced with a comprehensive pattern that handles OSC title sequences (`\x1b]0;...\x07`), DEC private modes (`\x1b[?25l`), and save/restore cursor (`\x1b7`/`\x1b8`). Added 3 new test cases.

### Minor
- **`_ensure_session` renamed** to `_session_exists`
- **`find_project_path` duplication** — noted for follow-up (bob's copy predates this PR)
- **tmpdir leak / parallel docstring** — valid points, will address in follow-up

All 2160 tests pass, lint clean.

PR #381, opened 2026-05-26
Bob runner: strip ANSI escapes on live-terminal write to stop flicker

### LINKED ISSUE #379: Bob runner: TUI escape sequences cause terminal to flash/flicker during streaming
## Summary

When running the factory with `--runner bob`, the terminal flashes/flickers rapidly during agent execution — visually disorienting (described as "epileptic"). This is **specific to the Bob runner**; the `claude` and `codex` runners stream cleanly.

## Root cause

Bob Shell is a TUI and emits ANSI/terminal control sequences (cursor-home, `\033[2J` clear-screen, alternate-screen toggles, spinner redraws) even in headless `-p` mode. The factory tees Bob's raw stdout/stderr straight into the parent terminal, so every redraw/clear sequence repaints the screen → flicker.

Two things combine:

1. **Live streaming passes raw bytes through.** `BobRunner.headless` enables streaming whenever stdout is a TTY:

   ```python
   # factory/runners/bob.py:245-259
   stream = should_stream()
   prefix = f"[bob:{role}]" if stream else None
   ...
   stream_subprocess(proc, stream=stream, prefix=prefix)
   ```

   `tee_stream` then does a line-buffered passthrough with no sanitization:

   ```python
   # factory/runners/_stream.py:44-53
   line = await src.readline()
   ...
   if stream:
       if prefix:
           dest.write(prefix)
       dest.write(line)
       dest.flush()
   ```

2. **The `[bob:<role>]` prefix is injected per line**, but Bob's escape sequences don't map cleanly to lines, so prefixes get interleaved mid-escape-sequence, which can corrupt/re-trigger redraws and make the flicker worse.

## Reproduction

```bash
factory ceo /path/to/project --runner bob
```

Observe rapid screen repainting/flashing while a Bob agent runs (in an interactive TTY).

## Workaround

Disable streaming — output is still fully captured and returned to the CEO, just not passed through live. `should_stream()` returns `False` if `FACTORY_RUNNER_QUIET=1` or stdout isn't a TTY:

```bash
FACTORY_RUNNER_QUIET=1 factory ceo /path/to/project --runner bob
```

Or in `~/.factory/config.toml`:

```toml
[defaults]
runner_quiet = "1"
```

## Proposed fix

Strip ANSI escape sequences in `tee_stream` before writing to `dest` (regex `\x1b\[[0-9;?]*[ -/]*[@-~]`), gated on the Bob runner so streaming stays useful but doesn't repaint the screen. Keep the captured buffer raw (returned to the CEO); only sanitize the bytes written to the live terminal.

Related files:
- `factory/runners/bob.py:245-259`
- `factory/runners/_stream.py:27-53`

colehurwitz:
Closes #379

## Changes

Bob Shell is a TUI that emits cursor/clear/alt-screen/spinner escape sequences even in headless `-p` mode. `tee_stream` wrote those raw to the parent terminal, so every Bob redraw repainted the parent screen — the rapid "epileptic" flicker reported in #379. The `claude` and `codex` runners stream line-oriented output and are unaffected.

The fix sanitizes **only** the bytes written to the live terminal, gated behind a `sanitize: bool = False` flag so `claude`/`codex` stay byte-identical and the captured buffer the CEO consumes stays fully **raw**.

- **`factory/runners/_stream.py`**: add module-level `strip_ansi()` + robust multi-branch byte regex covering CSI (ECMA-48 `[0-?]` params — colon-delimited truecolor & DEC private modes like `\x1b[?1049h`/`\x1b[?25l`), OSC (BEL/ST terminator), Fe/C1, and Fp (DECSC/DECRC/keypad). Preserves `\r`/`\n` and UTF-8 multibyte (8-bit C1 ST `\x9C` deliberately not matched). Thread `sanitize` through `stream_subprocess` → **both** `tee_stream` calls (stdout + stderr). `tee_stream` sanitizes only `dest` writes (buffer always raw) and skips empty-after-strip lines so redraw-only output doesn't flood the terminal with bare `[bob:role]` prefixes.
- **`factory/runners/bob.py`**: pass `sanitize=True` at the Bob call site only. `claude.py`/`codex.py` untouched.
- **Tests** (`tests/test_runners.py`, `tests/test_codex_runner.py`): ~12 tests mirroring `TestStreamingOutput` — `strip_ansi` units, sanitize-keeps-buffer-raw, prefix/redraw-skip, no-sanitize byte-identical, `stream_subprocess` threads to both streams, and runner wiring (bob sanitizes; claude/codex do not).

## Out of scope (per #379)
- Captured buffer / saved review file stay raw (the CEO consumes them).
- `cli.py`'s own colored output/spinner escapes are untouched.

## Verification
- `pytest tests/test_runners.py tests/test_codex_runner.py` — 86 passed
- Smoke suite (`test_models`, `test_guards`, `test_cli`) — 298 passed
- `ruff check` and `mypy factory/runners/` — clean

### DIFF
--- a/factory/runners/_stream.py
+++ b/factory/runners/_stream.py
@@ -8,11 +8,15 @@ import sys
 from typing import BinaryIO
 
 # Robust multi-branch matcher for ANSI/VT escape sequences. Deliberately
-# preserves \r and \n (they are not ANSI escapes). Covers four classes:
+# preserves \r and \n (they are not ANSI escapes). Covers five classes:
 #   - CSI: \x1b[ <params 0x30-0x3F> <intermediates 0x20-0x2F> <final 0x40-0x7E>
 #     (colors incl. colon-delimited truecolor, cursor moves, clear-screen,
 #     alt-screen toggles \x1b[?1049h/l, cursor-visibility \x1b[?25l/h)
 #   - OSC: \x1b] ... terminated by BEL (\x07) or ST (\x1b\\) — e.g. window title
+#   - String-introducer DCS/SOS/PM/APC: \x1bP, \x1bX, \x1b^, \x1b_ carry a payload
+#     terminated by BEL (\x07) or ST (\x1b\\). This branch MUST precede the Fe/C1
+#     branch so the whole payload is consumed — otherwise Fe greedily matches just
+#     the 2-byte introducer and the payload leaks as visible text.
 #   - Fe / 2-byte C1: \x1b followed by 0x40-0x5F (incl. ESC M reverse line feed)
 #   - Fp: \x1b7 (DECSC), \x1b8 (DECRC), \x1b= / \x1b> (keypad modes)
 # The 8-bit C1 ST (\x9C) is intentionally NOT matched: on a raw byte stream that
@@ -22,6 +26,7 @@ _ANSI_ESCAPE_RE = re.compile(
     rb"\x1B(?:"
     rb"\[[0-?]*[ -/]*[@-~]"              # CSI ... <final>
     rb"|\][^\x07\x1B]*(?:\x07|\x1B\\)"   # OSC ... (BEL or ST terminator)
+    rb"|[PX^_][^\x07\x1B]*(?:\x07|\x1B\\)"  # DCS/SOS/PM/APC ... (BEL or ST terminator)
     rb"|[@-Z\\-_]"                        # 2-byte C1 / Fe (incl. ESC M)
     rb"|[78=>]"                           # Fp: DECSC, DECRC, keypad =/>
     rb")"
--- a/factory/runners/_stream.py
+++ b/factory/runners/_stream.py
@@ -3,8 +3,34 @@
 from __future__ import annotations
 
 import asyncio
+import re
 import sys
 from typing import BinaryIO
+
+# Robust multi-branch matcher for ANSI/VT escape sequences. Deliberately
+# preserves \r and \n (they are not ANSI escapes). Covers four classes:
+#   - CSI: \x1b[ <params 0x30-0x3F> <intermediates 0x20-0x2F> <final 0x40-0x7E>
+#     (colors incl. colon-delimited truecolor, cursor moves, clear-screen,
+#     alt-screen toggles \x1b[?1049h/l, cursor-visibility \x1b[?25l/h)
+#   - OSC: \x1b] ... terminated by BEL (\x07) or ST (\x1b\\) — e.g. window title
+#   - Fe / 2-byte C1: \x1b followed by 0x40-0x5F (incl. ESC M reverse line feed)
+#   - Fp: \x1b7 (DECSC), \x1b8 (DECRC), \x1b= / \x1b> (keypad modes)
+# The 8-bit C1 ST (\x9C) is intentionally NOT matched: on a raw byte stream that
+# is later UTF-8 decoded, 0x9C is a valid continuation byte and matching it could
+# clip a multibyte character. A lone trailing \x1b is left as-is.
+_ANSI_ESCAPE_RE = re.compile(
+    rb"\x1B(?:"
+    rb"\[[0-?]*[ -/]*[@-~]"              # CSI ... <final>
+    rb"|\][^\x07\x1B]*(?:\x07|\x1B\\)"   # OSC ... (BEL or ST terminator)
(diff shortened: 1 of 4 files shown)


colehurwitz:
**Independent review: LGTM** — the fix is correct, well-scoped, and the regex holds up under empirical testing (CSI/OSC/DCS-SOS-PM-APC/Fe/Fp branches, branch ordering, ReDoS, UTF-8 preservation all verified). Local run: `pytest tests/test_runners.py tests/test_codex_runner.py` → **88 passed**.

## Strengths

- **Tight scope.** `claude.py` and `codex.py` call `stream_subprocess(...)` without `sanitize`, so they default to `False` and stay byte-identical. Only the Bob call site opts in. Verified by reading all three runners plus the two negative wiring tests.
- **Raw-buffer invariant is solid.** `buffer.append(line)` happens before any sanitization and is never touched again; `strip_ansi` only ever feeds `dest`. The dedicated `test_tee_stream_sanitize_strips_dest_keeps_buffer_raw` test pins this. The CEO consumes the raw buffer, so behavior is preserved.
- **Branch ordering is correct and intentional.** The DCS/SOS/PM/APC branch (`[PX^_]...`) precedes the Fe branch (`[@-Z\\-_]`), which is necessary because `P`/`X`/`^`/`_` fall inside the Fe char class — without the earlier branch, Fe would greedily match just the 2-byte introducer and leak the payload. The inline comment calls this out. Confirmed empirically: `\x1bP1$r0m\x1b\\after` → `after`.
- **UTF-8 safety.** The deliberate omission of 8-bit C1 ST (`\x9C`) avoids clipping multibyte continuation bytes. `strip_ansi("café — 日本語".encode())` round-trips unchanged.
- **Redraw-skip logic is precise.** `out != line and not out.strip(b"\r\n")` drops lines that became empty *because* escapes were stripped (incl. spinner frames like `\x1b[2K\r` → `\r` → dropped) while preserving genuine blank lines (`\n` → `out == line`, not dropped) and progress lines with text. All three cases are covered by tests.
- **No ReDoS.** Every alternative is linear with no nested quantifiers; negated char classes (`[^\x07\x1B]*`) bound the string branches. Stress tests (200k-byte unterminated OSC, 100k CSI params, 100k bare ESC) all complete in <2ms.
- **Honest, accurate documentation.** The module comment correctly documents the stateless/line-oriented limitation (unterminated or readline-split sequences may leak) rather than overclaiming. No secrets handling concerns — sanitization is purely subtractive on display bytes.

## Minor observations (non-blocking)

- **Known limitation — split/unterminated sequences (`_stream.py:25-30`, severity: low/informational).** Because stripping is per-`readline()` chunk, a sequence split across a line boundary, or an unterminated OSC/DCS, leaks its payload to the terminal (verified: `\x1b]0;title-no-term` → `0;title-no-term`; split CSI `hello\x1b[` + `2Jworld` leaks `2Jworld`). This is correctly documented as out of scope and is low-probability for Bob, whose escapes normally arrive intact per line. Worth keeping in mind if a future TUI emits chunkier output. No change requested.
- **`RIS` (`\x1bc`, full reset) is not stripped (severity: low).** `\x1bc` (ESC c) isn't in any branch — `c` (0x63) is above the Fe range (`@-Z`, `\-_`). Bob is unlikely to emit a full terminal reset in headless mode, and adding it risks little, but if reset-flicker ever shows up this is the gap. Informational only.

## Verdict

The implementation matches the issue's proposed fix, the invariants are enforced and tested, and the blast radius is contained to the Bob runner. Test coverage is thorough — units for each escape class, the buffer-raw invariant, redraw-skip vs. genuine-blank distinction, byte-identical no-sanitize path, threading to both stdout/stderr, and positive/negative runner wiring. No blocking issues. Ship it.

— Independent reviewer (fresh-context subagent)

colehurwitz:
@akashgit made some bob fixes. should be ready to merge.

colehurwitz:
@akashgit can you take a look at this. want to merge and test bob more

xukai92 (approved):
LGTM — narrowly scoped sanitization on the live-terminal write path only; raw bytes preserved in the captured buffer so downstream parsing is unaffected. Regex is compiled once at module level, terminator-aware, linear-time, no catastrophic backtracking, and gated behind `sanitize: bool = False` so Claude/Codex stay byte-identical. Test coverage spans the regex unit cases, `tee_stream` behavior, and per-runner wiring.

Approving.

PR #397, opened 2026-05-28
ci: rename plugin branch to plugins and document plugin install

xukai92:
## What
- rename the bundled Claude Code plugin workflow from `plugin` to `plugins`
- update README instructions for installing re:factory from the `plugins` branch
- keep publishing a legacy `plugin` branch mirror for backward compatibility

## Why
The branch and workflow name now match the user-facing `/plugin marketplace add akashgit/remote-factory#plugins` command while preserving existing installs that still reference `#plugin`.

## Split
This PR intentionally contains only the plugin branch/workflow/README changes. The phase 0 harness abstraction work is split into a separate PR.

### DIFF
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -1,4 +1,4 @@
-name: Build plugin branch
+name: Build plugins branch
 
 on:
   push:
@@ -30,12 +30,12 @@ jobs:
       - name: Copy skills to .agents/skills
         run: cp -r skills/ .agents/skills/
 
-      - name: Push to plugin branch
+      - name: Push to plugins branch
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git checkout -B plugin
+          git checkout -B plugins
           git add -f agents/ codex-agents/ .agents/skills/
           git diff --cached --quiet && echo "No changes" && exit 0
           git commit -m "build: generate plugin agent files from $(git rev-parse --short main)"
-          git push --force origin plugin
+          git push --force origin plugins
--- a/.github/workflows/plugins.yml
+++ b/.github/workflows/plugins.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Copy skills to .agents/skills
         run: cp -r skills/ .agents/skills/
 
-      - name: Push to plugins branch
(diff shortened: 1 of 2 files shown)


colehurwitz:
## Review

Docs + CI rename. Verified the README claims against the repo (manifests, `factory install` help, `skills/` contents) — everything lines up.

### Overview
- Renames `.github/workflows/plugin.yml` → `plugins.yml`, the workflow display name, and the published branch (`plugin` → `plugins`) so they match the `/plugin marketplace add akashgit/remote-factory#plugins` UX.
- Adds an **Install as a Claude Code Plugin** section to the README and reframes the existing **Plugin Agents** section as the "skip the marketplace" alternative, with a new `--runner codex` example.

### Correctness
- Workflow trigger (`push: branches: [main]`) matches the README's "rebuilds the `plugins` branch on every push to `main`". ✅
- `/plugin install factory@remote-factory` matches `.claude-plugin/marketplace.json` (`remote-factory` marketplace, `factory` plugin). ✅
- `/factory:implement` is the only slash command shipped (`commands/implement.md`). ✅
- `pipeline-subagents` and `implement` skills both exist under `skills/`. ✅
- `factory install --runner codex` → `~/.codex/agents/` matches `factory/cli.py:3724`. ✅
- Grep confirms no other docs/workflows reference the old `plugin` branch name, so no orphaned references after the rename.

### Nits (non-blocking)
- `plugins.yml` still has `jobs.build-plugin:` (singular) and the commit message string `"build: generate plugin agent files from …"`. Cosmetic only — `"plugin"` reads fine as an adjective there — but if you want full consistency, `jobs.build-plugins` would match the rename.
- First post-merge gap: until the next push to `main` actually fires `plugins.yml`, the new `plugins` branch won't exist on origin. Worth merging a trivial follow-up commit (or running the workflow manually) right after merge so the README instructions don't 404 for early adopters.
- The "Follow-up: delete the existing `plugin` branch manually" is fine, but anyone who already ran `/plugin marketplace add …#plugin` will silently get a frozen snapshot once `main` advances past the deletion. A one-liner in the new README section telling existing users to re-add with `#plugins` would be a nice touch — not required for this PR.

### Risk
Minimal. No runtime code paths touched; the only behavioral change is which remote branch the workflow force-pushes to.

**Verdict: LGTM** — ship it, optionally fold in the `jobs.build-plugins` rename before merge.

Issue #222, opened 2026-05-16
Finalize gate overrides KEEP to REVERT on pre-existing smoke test failures


akashgit:
## Problem

`factory finalize --verdict keep` runs an internal precheck that includes the smoke test. When the smoke test fails — even if the failure is pre-existing and also occurs on main — the finalize gate silently overrides the CEO's KEEP verdict to REVERT.

This has now happened **twice** in consecutive experiments:
- **Experiment 55** (PR #218): Finalize overrode KEEP → REVERT due to flaky `test_loop_logs_sleep_message`
- **Experiment 56** (PR #221): Finalize overrode KEEP → REVERT due to pytest hanging during teardown (exits code 2)

Both times, the CEO had to manually edit `results.tsv` and `verdict.json` to correct the verdict.

## Root Cause

`factory/cli.py` `cmd_finalize()` runs `precheck()` internally and if it fails, overrides the verdict:
```
Finalize gate: precheck FAILED — overriding keep to revert (smoke_test)
```

The precheck has no concept of "pre-existing failure" — it doesn't compare the smoke test result against main. If the smoke test fails on the PR branch, it fails the gate, even if main has the same failure.

## Expected Behavior

The finalize gate should either:
1. **Compare smoke test against main**: Run the smoke test on both branches and only fail if the PR branch introduces a NEW failure
2. **Accept `--force` flag**: Allow `factory finalize --verdict keep --force` to bypass precheck when the CEO has already verified the failure is pre-existing
3. **Use `--score-before` / `--score-after` properly**: When both scores are provided AND above threshold, the score_direction check should pass without re-running eval. Currently the finalize re-runs precheck even when scores are explicitly provided.

## Workaround

Manually edit `.factory/results.tsv` and `.factory/experiments/NNN/verdict.json` to change `revert` → `keep` and remove the `[OVERRIDDEN]` prefix from notes.

## Related

- Experiment 55: PR #218 (closed, superseded)
- Experiment 56: PR #221 (open)
- The smoke test command in `config.json` (`pytest tests/test_models.py tests/test_guards.py tests/test_cli.py -x -q --tb=short`) includes `test_cli.py` which has a hanging test that causes pytest to exit with code 2

akashgit:
Fixed in PR #232 (merged). Two changes:
1. Added `--force` flag to `factory finalize` — bypasses the internal precheck gate when the CEO has already verified failures are pre-existing
[...]

PR #417, opened 2026-06-01
fix: --force must not bypass hard constraints

RobotSail:
## Summary

- `--force` on `factory finalize` now only bypasses soft prechecks (smoke test, score direction, anti-pattern detection)
- Hard constraints from `factory.md` are **always enforced** regardless of `--force`
- Hard constraints run first, before any other precheck — if they fail, the verdict is overridden to revert even with `--force`

## Motivation

Hard constraints are user-defined invariants (e.g. GPU memory caps, quality gates) that the CEO should never circumvent. Since #232 added `--force` and the CEO prompt templates include it by default, hard constraints were effectively disabled — the CEO could `--force` past any constraint violation.

This was observed in practice: a factory running with a 24GB GPU memory constraint (`--gpu-memory-utilization 0.29`) had the CEO "fix" the value to 0.95 and keep the experiment using `--force`, completely bypassing the user's hardware constraint.

## What changed

`cmd_finalize` in `factory/cli.py`:
- Hard constraints are checked **outside** the `if not force` block — they always run
- Soft prechecks remain inside `if not force` — `--force` still bypasses them (preserving the fix for #222)
- `hard_constraints=None` is passed to `run_precheck` to avoid double-checking constraints

## Test plan

- [x] `test_force_flag_does_not_bypass_hard_constraints` — `--force` with failing hard constraint still reverts
- [x] `test_force_flag_bypasses_non_constraint_precheck` — `--force` with passing constraints skips soft precheck
- [x] `test_force_flag_with_revert_is_noop` — unchanged
- [x] All 23 hard constraint tests pass
- [x] Full test suite: 309 passed, 1 pre-existing failure (unrelated `test_interactive_task_contains_idea_text`)
- [x] `ruff check` clean

### DIFF
--- a/factory/cli.py
+++ b/factory/cli.py
@@ -937,7 +937,8 @@ def cmd_finalize(args: argparse.Namespace) -> int:
                 _emit_cli_event(project_path, "verdict.force_kept", {
                     "exp_id": args.id,
                 })
-                print("Finalize gate: precheck SKIPPED (--force), hard constraints PASSED")
+                hc_msg = ", hard constraints PASSED" if config.hard_constraints else ""
+                print(f"Finalize gate: precheck SKIPPED (--force){hc_msg}")
 
     cost = args.cost
     if cost is None:
--- a/factory/cli.py
+++ b/factory/cli.py
@@ -885,41 +885,59 @@ def cmd_finalize(args: argparse.Namespace) -> int:
 
     force = getattr(args, "force", False)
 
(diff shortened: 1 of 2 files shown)


xukai92:
@ceo-review

gx-ai-architect:
@ceo-review

RobotSail:
@ceo-review

osilkin98:
@ceo-review

PR #422, opened 2026-06-01
fix: write agent prompts to temp files to avoid ARG_MAX limit

Maxusmusti:
## Summary
- The CEO agent prompt (~153KB) was passed inline as a `--append-system-prompt` CLI argument, exceeding Linux's `ARG_MAX` per-argument limit and causing `[Errno 7] Argument list too long: 'claude'`
- Switched all three invocation sites (`headless`, `interactive_run`, `run_in_tmux`) to write the prompt to a temporary file and use `--append-system-prompt-file` instead
- Temp files are cleaned up in `finally` blocks after execution completes

## Test plan
- [x] All 2179 tests pass (`pytest -v`)
- [x] Verified `--append-system-prompt-file` flag works with `claude` CLI
- [x] Updated tests in `test_runners.py` and `test_cli.py` to match new flag

### DIFF
--- a/factory/runners/_tmux_persist.py
+++ b/factory/runners/_tmux_persist.py
@@ -83,7 +83,10 @@ async def run_in_tmux(
     exitcode_file = tmpdir / "exitcode"
     wrapper_script = tmpdir / "wrapper.sh"
 
-    cmd = ["claude", "--append-system-prompt", prompt]
+    prompt_file = tmpdir / "prompt.md"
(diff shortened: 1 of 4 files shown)


akashgit (approved):
CEO review: all 2179 tests pass, ruff clean, mypy clean. Temp file lifecycle is correct (write→close→use→unlink in finally). All 3 invocation sites converted consistently. LGTM.

PR #481, opened 2026-06-05
fix: skip test_available_runners_detected in CI

### LINKED ISSUE #479: bug: test_available_runners_detected fails in CI (no authenticated runners)
## Description

`tests/test_runner_e2e.py::test_available_runners_detected` fails across all Python versions (3.11, 3.12, 3.13) in CI because no runners are authenticated in the GitHub Actions environment.

```
AssertionError: No runners detected — auth detection may be broken
assert 0 > 0
```

This was introduced with the runner v2 changes and causes all CI test jobs to fail on every PR.

## Repro

Any PR will show the failure — e.g. https://github.com/akashgit/remote-factory/actions/runs/26991368105

## Possible fixes

- Skip the test when no runners are authenticated (`pytest.mark.skipif` or check env)
- Mark it as an integration test that only runs with explicit opt-in (e.g. `@pytest.mark.integration`)
- Mock the runner detection in CI

akashgit:
## Summary

- Skip `test_available_runners_detected` when `CI=true` — no runners are authenticated in GitHub Actions, so the test always fails
- Test still runs locally where runners are installed
- Closes #479

## Test plan

- [x] Test passes locally (no `CI` env var)
- [x] Test skips with `CI=true`
- [ ] CI should go green once merged

### DIFF
--- a/tests/test_runner_e2e.py
+++ b/tests/test_runner_e2e.py
@@ -350,6 +350,10 @@ def test_capability_matrix() -> None:
                 )
 
 
+@pytest.mark.skipif(
+    os.environ.get("CI") == "true",
+    reason="No runners authenticated in CI",
+)
 def test_available_runners_detected() -> None:
     """At least one runner is detected as available and authenticated."""
     assert len(AVAILABLE_RUNNERS) > 0, (


xukai92:
@ceo-review

PR #488, opened 2026-06-05
fix: fetch origin before creating worktree to prevent stale branches

### LINKED ISSUE #487: Worktree created from stale local branch, not origin
## Problem

`create_worktree()` in `factory/worktree.py` creates the worktree from the **local** base branch (e.g. `main`) without fetching from origin first. If `origin/main` has new commits (e.g. a recently merged PR), the worktree is stale from the moment it's created.

This causes the CEO to research, distill specs, and propose improvements for work that's already been done. The failure mode is silent and expensive: the entire research/distill cycle produces a confident, detailed improvement spec duplicating completed work. Only a human catching the mismatch prevents wasted builder/reviewer/evaluator cycles downstream.

## Reproduction

1. Have a project with a factory run that produced a PR
2. Merge the PR to `origin/main`
3. Start a new factory run before pulling locally
4. The CEO researches stale code and proposes re-doing the merged work

## Proposed Fix

In `create_worktree()`:
1. Run `git fetch origin <base_branch>` before creating the worktree
2. Use `origin/<base_branch>` as the start point instead of the local branch
3. Fall back to the local branch if there's no remote (local-only repos)
4. Log whether the fetch succeeded and how many commits behind the local branch was

This fixes it for all callers (`cmd_ceo`, `_run_single_cycle`, future entry points) in one place.

lukeinglis:
## Summary

- `create_worktree()` now runs `git fetch origin <base_branch>` before creating the worktree, branching from `origin/<base_branch>` instead of the (possibly stale) local branch
- Falls back gracefully to the local branch for repos without a remote
- Logs a warning with the commit delta when the local branch is behind origin

## Problem

When a PR is merged to `origin/main` but the user hasn't pulled locally, the factory creates a worktree from the stale local `main`. The CEO then researches outdated code, distills an improvement spec, and proposes re-doing work that was already completed in the merged PR. The failure mode is silent and expensive: the entire research/distill/strategy cycle is wasted.

Closes #487

## Test plan

- [x] 26/26 worktree tests pass (5 new)
- [x] `TestFetchAndResolveBase`: verifies `origin/main` is returned when behind, when up-to-date, and that local branch is returned when there's no remote
- [x] `TestCreateWorktreeWithRemote`: end-to-end test confirming the worktree contains commits from origin that the local branch doesn't have
- [x] `ruff check` and `mypy` clean

### DIFF
--- a/factory/worktree.py
+++ b/factory/worktree.py
@@ -17,12 +17,17 @@ def _fetch_and_resolve_base(project_path: Path, base_branch: str) -> str:
     ``origin/<base_branch>`` so the worktree starts from the remote tip.
     Falls back to the local ``<base_branch>`` for repos without a remote.
     """
-    fetch = subprocess.run(
-        ["git", "fetch", "origin", base_branch],
-        cwd=project_path,
-        capture_output=True,
-        text=True,
-    )
+    try:
+        fetch = subprocess.run(
+            ["git", "fetch", "origin", base_branch],
+            cwd=project_path,
+            capture_output=True,
+            text=True,
+            timeout=30,
+        )
+    except subprocess.TimeoutExpired:
+        log.warning("worktree_fetch_skipped", reason="timeout", base=base_branch)
+        return base_branch
     if fetch.returncode != 0:
         log.info("worktree_fetch_skipped", reason="no_remote", base=base_branch)
         return base_branch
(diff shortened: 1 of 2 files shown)


xukai92:
@ceo-review

Issue #369, opened 2026-05-25
Add factory handoff CLI command for cross-session resume


lukeinglis:
Factory experiment 5. Hypothesis: factory handoff CLI command for cross-session resume.

## What to Build

Create `factory/handoff.py` with a `generate_handoff(project_path: Path) -> str` function that reads 9 data sources and synthesizes them into a structured markdown brief printed to stdout. Add `cmd_handoff` to `factory/cli.py` following the `cmd_export` pattern. Register the `handoff` subparser with a single `path` argument.

### Data Sources
1. Git state (branch, recent commits)
2. Cycle state (.factory/state/cycle.json)
3. Checkpoint (.factory/checkpoint.json)
4. Config (.factory/config.json)
5. Strategy (.factory/strategy/current.md)
6. Backlog (.factory/strategy/backlog.md)
7. Experiment history (results.tsv)
8. Recent events (.factory/events.jsonl)
9. Review verdicts (.factory/reviews/)

### Output Format
Structured markdown brief to stdout with sections: Project, Current State, Score Trajectory, What's In Progress, What's Pending, Recent Activity, Recommended Next Steps.

## Acceptance Criteria
- [ ] `factory handoff /path` prints a structured markdown brief to stdout
- [ ] Missing files are handled gracefully (no crashes)
- [ ] Output includes mode, phase, branch, open PRs, last artifact, next steps, pending work
- [ ] At least 2 unit tests
- [ ] All tests pass
- [ ] Eval score does not regress

## Constraints
- Read CLAUDE.md before starting
- Do NOT touch files outside declared scope
- Do NOT modify eval/score.py or .factory/

lukeinglis:
Duplicate of #352. Created in error during experiment pipeline — should have used the existing issue.

Issue #352, opened 2026-05-25
Add factory handoff CLI command for cross-session resume


lukeinglis:
Factory experiment 3. Hypothesis: factory handoff CLI command for cross-session resume.

## What to Build

Add a `factory handoff <path>` CLI command that synthesizes the current `.factory/` state into a single actionable markdown brief, printed to stdout.

### Implementation

1. **New function `cmd_handoff` in `factory/cli.py`**:
   - Accept a project path argument (positional, like other commands)
   - Read and synthesize these sources:
     - `.factory/state/cycle.json` (mode, respawns, cycle_id)
     - `.factory/checkpoint.json` (completed/pending agents)
     - `.factory/events.jsonl` (latest events, find current phase)
     - `.factory/strategy/current.md` (approved hypotheses)
     - `.factory/strategy/backlog.md` (remaining work)
     - `.factory/reviews/*-latest.md` (latest agent outputs)
     - Git state: `git branch --show-current`, `git log --oneline -3`
     - Open PRs: `gh pr list --state open` (if available, gracefully handle failure)
   - Print a structured markdown brief to stdout:
     ```
     # Factory Handoff — <project name> @ <timestamp>

     **Mode:** <from cycle.json or events>
     **Phase:** <current phase from events/checkpoint>
     **Branch:** <current git branch>
     **Open PRs:** <list or "none">
     **Focus:** <if targeted mode, show focus target>

     ## Last Successful Artifact
     <latest review file or strategy file, with path>

     ## Recommended Next Steps
     1. <specific action based on current phase>
     2. <specific action>

     ## Key Files
     - Strategy: .factory/strategy/current.md
     - Research: .factory/strategy/research.md
     - Backlog: .factory/strategy/backlog.md
     - Reviews: .factory/reviews/

     ## Pending Work
     <from backlog.md or remaining hypotheses in current.md>
     ```
   - Handle missing files gracefully (any `.factory/` file may not exist)

2. **Register the subparser in `build_parser()`**:
   - Add `handoff` subcommand with project path as positional argument
   - Wire to `cmd_handoff` handler

3. **Tests in `tests/test_cli.py`**:
   - Test with a minimal `.factory/` directory (just config.json)
   - Test with full state (cycle.json, events.jsonl, strategy, reviews)
   - Test with missing files (graceful degradation)

## Acceptance Criteria
- [ ] `factory handoff /path` prints a structured markdown brief to stdout
- [ ] Missing files are handled gracefully (no crashes)
[...]

PR #493, opened 2026-06-05
feat: code-level issue dedup via factory issue-resolve

lukeinglis:
## Summary

Alternative approach to #375 for fixing #372 (CEO creates duplicate GitHub issues). Moves the dedup logic from the CEO prompt into tested Python code.

- Add `extract_issue_refs()` to parse hypothesis text for issue references with priority ordering (explicit `Addresses #N` > `**Backlog item:** #N` > `issue #N` > single bare `#N`)
- Add `resolve_reusable_issue()` to verify the referenced issue is open and relevant (keyword match against title)
- Add `factory issue-resolve` CLI command the CEO can call before `gh issue create`
- 14 new tests covering parsing, resolution, and edge cases (ambiguous refs, closed issues, irrelevant titles)

## How the CEO uses it

The 37-line Issue Reuse Protocol in PR #375's `ceo.md` shrinks to:

```
EXISTING_ISSUE=$(factory issue-resolve "$PROJECT_PATH" --hypothesis "<hypothesis text>")
if [ -n "$EXISTING_ISSUE" ]; then
  ISSUE_NUM=$EXISTING_ISSUE
else
  ISSUE_NUM=$(gh issue create ...)
fi
```

## Why code over prompt

- All regex parsing, open-state checks, and relevance validation are unit-tested
- Works identically across all modes (Improve step 2c, Research R3a, Refine R3) without duplicating instructions
- Less likely to regress when the prompt is edited for unrelated reasons
- Edge cases (ambiguous bare refs, closed issues, poisoned references) are handled deterministically

## Test plan

- [x] 54/54 issue tests pass (14 new)
- [x] `TestExtractIssueRefs`: explicit refs, bare refs, ambiguous multi-ref, path fragments
- [x] `TestResolveReusableIssue`: open+relevant, closed, irrelevant title, no refs
- [x] `ruff check` and `mypy` clean

### DIFF
--- a/factory/cli.py
+++ b/factory/cli.py
@@ -857,6 +857,17 @@ def cmd_guard(args: argparse.Namespace) -> int:
     return 0
 
 
+def cmd_issue_resolve(args: argparse.Namespace) -> int:
+    """Check if a hypothesis references a reusable open GitHub issue."""
+    from factory.issue import resolve_reusable_issue
+
+    project_path = Path(args.path)
+    number = resolve_reusable_issue(args.hypothesis, project_path)
+    if number is not None:
+        print(number)
+    return 0
+
+
 def cmd_begin(args: argparse.Namespace) -> int:
     from factory.store import ExperimentStore
 
@@ -3702,6 +3713,12 @@ def build_parser() -> argparse.ArgumentParser:
                     help="Also check fixed surface constraints (research mode)")
 
     # begin
+    p = sub.add_parser("issue-resolve",
+                        help="Check if hypothesis references a reusable open issue")
+    p.add_argument("path", help="Path to the project")
+    p.add_argument("--hypothesis", required=True, help="Hypothesis text to scan for issue refs")
+
+    # begin
     p = sub.add_parser("begin", help="Start experiment, print ID")
     p.add_argument("path", help="Path to the project")
     p.add_argument("--hypothesis", required=True, help="Experiment hypothesis text")
@@ -4209,6 +4226,7 @@ def main(argv: list[str] | None = None) -> int:
         "init": cmd_init,
         "eval": cmd_eval,
         "guard": cmd_guard,
+        "issue-resolve": cmd_issue_resolve,
         "begin": cmd_begin,
         "finalize": cmd_finalize,
         "history": cmd_history,
--- a/factory/issue.py
+++ b/factory/issue.py
@@ -175,6 +175,78 @@ def is_issue_ref(ref: str) -> bool:
     return False
 
 
+_ISSUE_REF_PATTERNS = [
+    re.compile(r"[Aa]ddresses:?\s*#(\d+)"),
+    re.compile(r"\*\*Backlog item:\*\*\s*#(\d+)"),
+    re.compile(r"[Ii]ssue\s+#(\d+)"),
+]
+_BARE_REF = re.compile(r"(?<![/\w])#(\d+)\b")
+
+
+def extract_issue_refs(text: str) -> list[int]:
+    """Extract issue numbers from hypothesis or backlog text.
+
+    Priority order:
+    1. Explicit references: ``Addresses #NNN``, ``**Backlog item:** #NNN``, ``issue #NNN``
+    2. Bare ``#NNN`` only if exactly one match and no explicit match found
+    """
+    explicit: list[int] = []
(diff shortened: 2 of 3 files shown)


osilkin98:
Context for the review below: I've been testing the Claude Fable model for triaging the open PR backlog. It flagged the following on this PR, and this CEO review session independently verified or refuted each claim against the current branch head. Process context: #517.

1. This does not fix the problem in issue #372: the canonical duplicate pair there (#352 vs #369) contains NO explicit issue references in the duplicate's text, so extract_issue_refs() returns nothing and a duplicate is created anyway. Real CEO duplicates are same-idea-different-words, which reference parsing cannot catch.
2. The tool is never invoked: ceo.md has 4 `gh issue create` sites and none call `factory issue-resolve` — the PR touches zero prompt lines, so on merge this is dead code.
3. The relevance check (any shared 4+ letter word between hypothesis and issue title) is loose enough to reuse the WRONG issue — which later gets closed by an unrelated experiment. Quiet compounding failure.
4. `gh issue list --search "<keywords>"` would address the actual same-idea-different-words case in a few lines and was not considered.

osilkin98 (changes requested):
## ❌ Factory Review: REVERT

**Verdict:** REVERT
**Reason:** Dead code on merge (no caller in any agent prompt), does not address the canonical duplicate scenario from #372 (semantic duplicates without explicit refs), and the relevance check has a high false-positive rate that can cause worse outcomes than the original bug.


### Precheck Gate

```
4 claims adjudicated: 4 confirmed, 0 refuted. 0 prior verdicts (none exist).
```

### Code Review Notes

- Claims and provenance are in my comment above. Each note below is this review's independent adjudication of one claim.
- CLAIM 1 CONFIRMED: extract_issue_refs returns nothing for the canonical #372 scenario because #369's hypothesis text contains no issue refs. Evidence: gh issue view 369 body contains zero #NNN references. Why it matters: the PR claims to fix #372, but the actual duplicate scenario (same idea, different words, no explicit refs) is not addressed.
- CLAIM 2 CONFIRMED: ceo.md has 4 gh-issue-create sites (lines 1149, 1170, 2063, 2479) and none call factory issue-resolve. The PR modifies zero prompt files. Evidence: grep -n issue-resolve factory/agents/prompts/ceo.md returns nothing. Why it matters: on merge this is dead code with no caller, so the feature cannot activate.
- CLAIM 3 CONFIRMED: the relevance check at factory/issue.py:238 matches any 4+ letter word shared between hypothesis and issue title. Tested 5 unrelated pairs and all produce false positives via common domain words (factory, data, dashboard, cache, http). Evidence: factory/issue.py:238-239. Why it matters: a false reuse is worse than a duplicate. It attaches experiment metadata and close-on-merge logic to the wrong issue, quietly corrupting issue tracking.
- CLAIM 4 CONFIRMED: gh issue list --search would match existing issues by title similarity without needing #NNN refs, directly solving #372's canonical scenario. The PR does not mention or consider this approach. Evidence: the canonical pair #352/#369 shares the title 'Add factory handoff CLI command for cross-session resume' verbatim. Why it matters: a search-based approach solves the hard case (semantic duplicates) that regex parsing cannot.
- ADDITIONAL: redundant subprocess call. resolve_reusable_issue calls fetch_issue (which runs gh issue view with title,body,labels,number,url) then runs a second gh issue view just for state. The state field could be included in the first call. Evidence: factory/issue.py:220 and 226-231. Why it matters: minor inefficiency, but indicates the code was bolted on rather than integrated with existing abstractions.
- ADDITIONAL: Codecov reports 75% patch coverage (11 lines missing). factory/cli.py has 40% patch coverage. For a feature whose value proposition is 'unit-tested deterministic behavior' vs the prompt approach, significant uncovered paths undermine that argument. Evidence: Codecov PR comment.
- Acceptance path to flip this to KEEP: (1) Add a ceo.md change that wires factory issue-resolve into all 4 gh-issue-create sites, so it is not dead code. (2) Replace the regex-only approach with gh issue list --search to catch semantic duplicates (the actual #372 scenario). (3) Tighten the relevance check to require multiple keyword matches or use a higher word-length threshold, eliminating false positives from common domain words like 'factory'. (4) Merge the state check into the existing fetch_issue call to avoid the redundant subprocess.

---
*Posted by Factory CEO*

RobotSail:
@lukeinglis can you reconcile this one with #375

PR #501, opened 2026-06-06
fix: isolate system prompt via AGENTS.md for Codex and OpenCode

### LINKED ISSUE #471: System prompt exposure in non-Claude runners (Codex, Bob, OpenCode)
## Problem

Codex, Bob, and OpenCode concatenate the system prompt and task into a single user-visible message. Unlike Claude's `--append-system-prompt-file`, this dumps the full factory CEO prompt (2000+ lines) at the start of every interactive session.

## Suggested Mitigations

- **Codex**: Write prompt to temporary `AGENTS.md` in the project dir — Codex reads it as system instructions
- **OpenCode**: Create temporary agent config with system prompt
- **Bob**: No known mechanism — concatenation may be the only option

## Context

From PR #443 consolidated review. Non-breaking UX issue — all runners function correctly, but the experience is degraded.

akashgit:
## Summary

Fixes #471 — system prompt exposure in non-Claude runners.

- **Codex**: System prompt is now written to a temporary `AGENTS.md` in the project directory. Codex auto-reads it as system-level instructions. Only the task is passed as the user message. Original `AGENTS.md` is backed up and restored in a `finally` block.
- **OpenCode**: Same approach — OpenCode also reads `AGENTS.md` for system instructions. Prompt goes to `AGENTS.md`, only the task is passed via `-p`.
- **Bob**: Left as concatenation — no known mechanism for separate system prompt delivery.

### Before
All non-Claude runners dumped the full 2000+ line factory CEO prompt at the start of every session as part of the user message.

### After
The system prompt is invisible to the user (loaded via `AGENTS.md`). Only the task appears as the user message.

## Test plan

- [x] `uv run pytest tests/test_codex_runner.py tests/test_runners.py -v` — 99 passed
- [x] `uv run pytest tests/ -x -q --tb=short -m "not slow"` — 2211 passed, 3 skipped
- [x] `uv run ruff check` — all checks passed
- [x] AGENTS.md lifecycle tests: creation, append to existing, cleanup, restore

### DIFF
--- a/factory/runners/_agents_md.py
+++ b/factory/runners/_agents_md.py
@@ -2,16 +2,17 @@
 
 Codex and OpenCode read AGENTS.md as system-level instructions. These helpers
 write the factory prompt into AGENTS.md (backing up any existing content) and
-restore it in a finally block, guarded by a file lock to prevent races.
+restore it in a finally block.
+
+Only used for interactive_run() — headless mode concatenates prompt+task inline.
 """
 
 from __future__ import annotations
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from pathlib import Path
 
 import structlog
-from filelock import FileLock
 
 log = structlog.get_logger()
 
@@ -23,7 +24,6 @@ class AgentsMdState:
     """Tracks the AGENTS.md file state for later restoration."""
 
     path: Path
-    lock: FileLock = field(repr=False)
     backup: str | None = None
 
 
@@ -34,34 +34,26 @@ def setup_agents_md(cwd: Path, prompt: str) -> AgentsMdState:
     crashed run — discard it (no backup). Otherwise back up and prepend.
     """
     agents_path = cwd / "AGENTS.md"
-    lock_path = cwd / ".factory" / ".agents_md.lock"
-    lock_path.parent.mkdir(parents=True, exist_ok=True)
-    lock = FileLock(lock_path)
-    lock.acquire()
 
     backup: str | None = None
     content = f"{SENTINEL}\n{prompt}\n"
 
-    try:
-        if agents_path.is_file():
-            existing = agents_path.read_text(encoding="utf-8")
-            if existing.startswith(SENTINEL):
-                log.debug("agents_md_stale_discarded", path=str(agents_path))
-            else:
-                backup = existing
-                content = f"{backup}\n{SENTINEL}\n{prompt}\n"
+    if agents_path.is_file():
+        existing = agents_path.read_text(encoding="utf-8")
+        if existing.startswith(SENTINEL):
+            log.debug("agents_md_stale_discarded", path=str(agents_path))
+        else:
+            backup = existing
+            content = f"{backup}\n{SENTINEL}\n{prompt}\n"
 
-        agents_path.write_text(content, encoding="utf-8")
-        log.debug("agents_md_written", path=str(agents_path), has_backup=backup is not None)
-    except BaseException:
-        lock.release()
-        raise
+    agents_path.write_text(content, encoding="utf-8")
+    log.debug("agents_md_written", path=str(agents_path), has_backup=backup is not None)
 
-    return AgentsMdState(path=agents_path, backup=backup, lock=lock)
+    return AgentsMdState(path=agents_path, backup=backup)
 
 
 def restore_agents_md(state: AgentsMdState | None) -> None:
-    """Restore AGENTS.md to its pre-setup state and release the lock."""
+    """Restore AGENTS.md to its pre-setup state."""
     if state is None:
         return
     try:
@@ -73,5 +65,3 @@ def restore_agents_md(state: AgentsMdState | None) -> None:
             log.debug("agents_md_removed", path=str(state.path))
     except OSError:
         log.debug("agents_md_restore_failed", path=str(state.path), exc_info=True)
-    finally:
-        state.lock.release()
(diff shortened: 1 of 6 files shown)


gx-ai-architect:
@ceo review

gx-ai-architect:
@ceo-review

gx-ai-architect:
@ceo-review redo the review, while also share the test you have run if you were to approve

gx-ai-architect:
@ceo-review

osilkin98 (dismissed):
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** PR correctly solves issue #471 (system prompt exposure in non-Claude runners) with solid test coverage and no regressions; all three prior claims are confirmed as real design gaps but none are blocking — they are hardening follow-ups, not merge blockers.

**Hypothesis:** Isolate system prompt via AGENTS.md for Codex and OpenCode runners (issue #471)

### Precheck Gate

```
3 claims adjudicated: 3 confirmed (all non-blocking). 2 additional findings.
```

### Code Review Notes

- These numbered claims came from a pre-review triage pass using the Claude Fable model. This CEO review independently verified or refuted each one against the code. Proposed as standard practice in issue #517.
- CLAIM 1 CONFIRMED (non-blocking): Double-backup poisoning on hard process death. If the process is SIGKILLed, OOM-killed, or tmux-killed after _setup_agents_md writes AGENTS.md but before _restore_agents_md runs, the factory-written content remains on disk. The next run treats it as the user's original and backs it up, permanently destroying the real file. Verified: grep confirms zero sentinel markers in written content, zero startup recovery paths. The fix commit (fe45e96) narrowed the exception-gap (moved write adjacent to try-finally from being 5 lines away in build_command), but does NOT address SIGKILL — Python finally blocks do not run on SIGKILL. Why it matters: a user's project AGENTS.md could be silently replaced with a 2000-line factory CEO prompt after a hard crash. Mitigation suggestion: add a sentinel comment (e.g. '# factory:system-prompt-start') and strip on startup.
- CLAIM 2 CONFIRMED (non-blocking): Concurrent factory runs race on AGENTS.md. The backup lives in instance state (self._agents_md_backup), and get_runner() at factory/runners/__init__.py:70 creates a fresh instance per invoke_agent() call. Two separate 'factory ceo' processes targeting the same project directory would race on read-write of the same AGENTS.md. Verified: no file-level locking exists around AGENTS.md (filelock is only used in store.py for experiment IDs and TSV writes). Why it matters: in practice, running two CEO processes against the same directory is not a supported use case and the factory already lacks concurrency guards at every other level. This is a pre-existing architectural gap, not a regression from this PR.
- CLAIM 3 CONFIRMED (non-blocking): _setup_agents_md and _restore_agents_md are copy-pasted between codex.py and opencode.py. Verified by diff: _setup_agents_md is byte-identical across both files. _restore_agents_md differs only in docstring noun, log event name, and one inconsistency — codex.py:190 uses getattr(self, '_agents_md_path', None) while opencode.py:148 uses self._agents_md_path directly. Both work because class-level annotations initialize the attributes, but the defensive-vs-direct inconsistency is a code smell. Why it matters: any future bug fix must land in two places. Extract to a shared mixin or factory/runners/_agents_md.py helper.
- ADDITIONAL: _setup_agents_md() is called OUTSIDE the try block in all 4 methods (codex.py headless:162 before try:163, codex.py interactive:213 before try:214, opencode.py headless:185 before try:186, opencode.py interactive:205 before try:206). If _setup_agents_md itself raises (e.g. PermissionError), _restore_agents_md will not run. Why it matters: low practical impact since a failed write leaves inconsistent backup state anyway, but moving the call inside the try block would be strictly more correct.
- ADDITIONAL: No test regressions. 2206 tests pass, 3 skipped. 5 failures in test_cli.py (TestCmdCeoInteractive, TestCmdCeoResearchIdeation) are pre-existing on main, verified by checkout. Lint (ruff) and type checks (mypy) clean on all changed files.

---
*Posted by Factory CEO*

osilkin98:
For reference, same setup as #514: I ran the open PR backlog through the Claude Fable model as a triage pass, and it flagged 3 issues on this PR. The CEO review above adjudicated each one independently (confirmed all 3, judged them non-blocking). The original claims it was given, verbatim:

1. Double-backup poisoning: if the process dies without running finally (SIGKILL, tmux kill, OOM), the factory-written AGENTS.md remains in the user's repo; the NEXT run's _setup_agents_md treats it as the user's original and backs it up, permanently destroying the real file. There is no sentinel marker identifying factory-written AGENTS.md and no startup recovery path.
2. Concurrent factory runs against the same project race on backup/restore: the backup lives in instance state (self._agents_md_backup), so two runners can clobber each other's view of the "original".
3. _setup_agents_md and _restore_agents_md are copy-pasted verbatim between codex.py and opencode.py, so any fix must land twice; extract shared helpers.

Follow-up work is tracked in #518. Process proposal for making this a standard part of CEO review is #517.

osilkin98:
One correction to my review above. The CEO detailed review from Jun 7 in this thread rated the AGENTS.md race as critical, with a concrete trigger path: `invoke_agents_parallel()` runs agents concurrently within a single factory run, so this doesn't need two separate factory processes to happen. My review called it non-blocking on the assumption that concurrent runs against one project aren't a supported use case, and that finding undercuts the assumption. I've updated #518 to track the fix (file locking around AGENTS.md setup/restore; `filelock` is already a dependency). If folks think this makes it a merge blocker rather than a follow-up, no objection from me.

gx-ai-architect:
@ceo-review

gx-ai-architect:
@ceo-review

gx-ai-architect:
@ceo-review

PR #491, opened 2026-06-05
Add BUILD_MODE fast/full/compile and disk space awareness to build-root

gx-ai-architect:
Closes #96

## Changes

- **build.sh**: Added `BUILD_MODE` env var (`fast`/`full`/`compile`) controlling verification depth, `PODMAN_STORAGE_ROOT` for custom Podman storage on machines with limited root disk, `PROJECT_SOURCE` volume mount for external project source, disk space check at startup (warns if < 20GB free), `fast_build()` and `clean_build()` helper functions, and `--layers` flag for cache-efficient image builds
- **build-root-ceo.md**: Added "Build Modes" section documenting when to use each mode, updated the auto-research loop EVALUATE step to use `BUILD_MODE=fast` during iteration and `BUILD_MODE=full` for final stage verification, added disk space management guidance
- **docs/build-root.md**: Added "Storage Configuration" section (disk budget, `PODMAN_STORAGE_ROOT` usage, threshold warnings) and "Build Modes" section (comparison table, usage examples), updated environment variables table with new vars
- **scripts/check-prerequisites.sh**: Added disk space verification that checks the target storage path (respects `PODMAN_STORAGE_ROOT`) and reports available space in the checklist
- **factory/build_root/stage.py**: Added `BuildMode` Literal type (`fast | full | compile`), added `last_build_mode` field to `PipelineStatus` model
- **tests/test_container_build.py**: New integration test file with Podman-gated container build tests (`test_containerfile_builds`, `test_fast_build_mini_project`), disk space check tests, and build mode env var routing tests (8 unit tests pass without Podman)

## Test plan

- [x] All 406 existing tests pass (`pytest tests/test_stage_detection.py tests/test_known_fixes.py tests/test_gradle_parser.py tests/test_cli.py tests/test_models.py tests/test_store.py -x -q`)
- [x] 8 new non-container tests pass
- [x] Ruff lint clean
- [ ] Container integration tests require Podman (skipped automatically if unavailable)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #516, opened 2026-06-10
Build root: Java equivalence testing tool + methodology docs

gx-ai-architect:
First PR enabling a basic Java build-root factory flow. Adds equivalence testing tool and methodology docs. Proven on Spring Framework v3.0.0.RELEASE (3846 classes, 15 modules, 14/15 API match) and v4.3.30.RELEASE (16776 tests, 0 failures).

### DIFF
--- a/docs/build-root-methodology.md
+++ b/docs/build-root-methodology.md
@@ -64,7 +64,8 @@ Before writing any code, answer these questions:
 **1. Containerfile**
 - Use the exact JDK version the project was built with, not just a compatible one
 - Install the exact compiler if it differs from javac (e.g., Eclipse JDT)
-- Set locale, timezone, encoding to match the original build environment
+- Set locale and encoding to match the original build environment (affects string formatting and character handling in code and tests)
+- Do NOT fake timestamps or system clock — the output should honestly say "built in 2026 from historical source," not pretend to be the original artifact
 - Pre-download build tool distributions (Gradle wrapper, Ant) to avoid download failures
 
 **2. Build system configuration**
--- a/docs/build-root-methodology.md
+++ b/docs/build-root-methodology.md
@@ -1,0 +1,174 @@
+# Build Root: Methodology for Rebuilding Ancient Software Packages
+
+## What Is a Build Root?
+
+A build root is a self-contained, containerized environment that can compile a historical software project from unmodified source code — even when the original build infrastructure no longer exists. The source code is never modified; all build behavior changes go through build system configuration, dependency remapping, and container setup.
+
+The purpose: enable CVE backporting to old, unsupported versions. You can't patch a vulnerability in code you can't compile.
+
+## The Hindsight Process
+
+After building Spring Framework v4.3.30.RELEASE (Gradle, 2020) and v3.0.0.RELEASE (Ant+Ivy, 2009) from source, here's the idealized process distilled from what actually worked — and what we had to learn the hard way.
+
+### Phase 1: Infrastructure Audit
+
+Before writing any code, answer these questions:
+
+**1. What build system does the project use?**
+- Gradle: you have init scripts (can intercept without touching source)
+- Maven: you have settings.xml and profiles (similar interception)
+- Ant + Ivy: no interception mechanism — must rewrite config files
+- Make/CMake: must modify Makefiles or provide toolchain files
+- The build system determines your entire fix strategy
+
+**2. What are ALL the dependency sources?**
+- List every repository URL in every build config file
+- Check each one with `curl -sI <url>` — which return 200, which return 401/403/404?
+- This is the single most important step. One dead repository can block the entire build.
+
+**3. What JDK/compiler/toolchain version is required?**
+- Check the project's documentation, CI configs, and compiler settings
+- Search for Docker images of that exact toolchain version
+- Don't assume alternatives are equivalent — we found behavioral differences between Sun JDK 6 and Zulu OpenJDK 6 that caused test failures
+
+**4. Does the project have internal build infrastructure?**
+- Some projects (like Spring) have build support modules that are separate repos or submodules
+- Check if those repos/submodules still exist
+- If they're dead (SVN externals, deleted repos), find the nearest surviving snapshot
+
+### Phase 2: Dependency Mapping
+
+**1. Extract every dependency declaration**
+- For each module, list every dependency with exact coordinates and version
+- For Maven/Gradle: `group:artifact:version`
+- For Ivy: `org/name/rev`
+- For non-standard naming (EBR, OSGi bundles): document the naming scheme
+
+**2. Verify each dependency on public repositories**
+- Check Maven Central: `curl -sI https://repo1.maven.org/maven2/<group-path>/<artifact>/<version>/<artifact>-<version>.jar`
+- Check other public repos: Gradle Plugin Portal, Red Hat Maven GA, JBoss
+- Classify each dependency: AVAILABLE, AVAILABLE_DIFFERENT_COORDS, UNAVAILABLE
+
+**3. For unavailable dependencies, exhaust all options in order**
+1. Search for the artifact under different coordinates on Maven Central
+2. Search alternative public repositories (Red Hat, JBoss, Eclipse)
+3. Search for open-source forks or reimplementations
+4. Build from archived source if available
+5. Create stub JARs with matching API signatures (last resort)
+6. Exclude the dependent module with documented justification (absolute last resort)
+
+**Key lesson learned:** We initially classified JMXMP (15 test failures) and Eclipse JDT compiler as "unavailable proprietary artifacts." Both turned out to be on Maven Central under different coordinates. Always search exhaustively before giving up.
+
+### Phase 3: Build Environment Construction
+
+**1. Containerfile**
+- Use the exact JDK version the project was built with, not just a compatible one
+- Install the exact compiler if it differs from javac (e.g., Eclipse JDT)
+- Set locale, timezone, encoding to match the original build environment
+- Pre-download build tool distributions (Gradle wrapper, Ant) to avoid download failures
+
+**2. Build system configuration**
+- For Gradle: init scripts in `~/.gradle/init.d/` (not `-I` flag — doesn't cover buildSrc)
+- For Ant+Ivy: replace ivysettings.xml, rewrite ivy.xml files via automated script
+- For Maven: settings.xml with mirror and profile overrides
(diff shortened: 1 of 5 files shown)


gx-ai-architect:
# Equivalence Testing: What It Actually Does and Doesn't Do

## Our Tool Does NOT Decompile

The current tool uses `javap`, which is a **disassembler** — it reads class file metadata (method signatures, field types, class hierarchy). It does NOT look at method body bytecode. This is an important distinction from full decompilation.

## Three Levels of Comparison

### Level 1: `jar tf` (class inventory)
- No decompilation. Lists file names inside the JAR.
- Works on any JAR, any JDK, any compiler.
- Zero dependency on how the artifact was built.
- **Our tool does this.**

### Level 2: `javap -public` (API surface)
- No decompilation. Reads the class file's constant pool and method/field descriptors, standardized in the JVM spec (section 4).
- You can run `javap` from JDK 21 on a class compiled by JDK 6 — works because the class file format is backward compatible.
- Does NOT depend on which compiler or builder produced the file.
- **Our tool does this.**

### Level 3: Full decompilation (CFR, Procyon, fernflower)
- Reads the actual bytecode of every method body and reconstructs Java source.
- **Compiler-dependent:** Sun javac, Eclipse JDT, and Zulu javac produce different bytecode patterns for the same source. The decompiler may reconstruct different-looking Java code from each.
- **Version-sensitive:** Decompilers are tuned for specific bytecode patterns. A decompiler written for Java 8 bytecode may produce messy output for Java 5 bytecode.
- **Noisy:** Variable names are lost (unless debug info is present), loop structures may differ, switch statements may be reconstructed differently. Tons of false diffs.
- **Our tool does NOT do this yet.**

## What Each Level Catches

| What could differ | Level 1 (inventory) | Level 2 (javap) | Level 3 (decompile) |
|---|---|---|---|
| Missing/extra class files | Catches | Catches | Catches |
| Changed method signature | No | Catches | Catches |
| Changed return type | No | Catches | Catches |
| Changed class hierarchy | No | Catches | Catches |
| Different method implementation (same signature) | No | No | Catches |
| Different algorithm/logic | No | No | Partially |
| Different dependency wiring at compile time | No | No | Partially |

## The Gap

The gap in our current tool: two methods with identical signatures can have completely different implementations, and levels 1-2 won't catch it.

For the build root use case, this risk is low because we compile the same source with the same compiler (Eclipse JDT 3.3.0) and same JDK (Sun 1.6.0_45). But it's not zero — especially where dependencies differ. Our Maven Central JARs vs the original EBR-repackaged JARs on the classpath could cause different constant folding or inlining by the compiler.

## Does Decompilation Require Matching the Original Build Tools?

**No.** The decompiler (CFR, Procyon, etc.) reads the standardized .class file format. You can decompile a class built by any compiler with any decompiler. The decompiler itself does not need to match the original build tools.

However, the **output quality** depends on how well the decompiler understands the bytecode patterns. Different compilers produce different bytecode for the same source:
- Sun javac generates one pattern for a for-each loop
- Eclipse JDT generates a slightly different pattern
- The decompiler reconstructs different-looking Java from each, even though the original source was identical

This means comparing decompiled output between original and rebuilt JARs will have **noise** from compiler pattern differences — even when there are no real behavioral differences. Noise filtering is needed.

## What Our Tool Currently Proves

**Binary compatibility:** Any code compiled against the original JAR will also work with our rebuild. All public methods, fields, constructors, and class hierarchies are identical. This is the guarantee that matters for drop-in replacement.

**What it does NOT prove:** Behavioral equivalence. Our methods could theoretically do something different internally (different bytecode, different optimization, different control flow). Level 3 decompilation would narrow this gap. Runtime testing (running an actual application against both JARs) would close it.

## Recommendation

1. **Current tool (levels 1-2) is useful** for proving binary compatibility and catching build errors (missing classes, wrong dependency versions changing API signatures).
2. **Level 3 should be added** for deeper verification — using CFR (single JAR, handles Java 5/6 well, available on Maven Central) with noise filtering to reduce false positives from compiler pattern differences.
3. **Runtime testing against a real application** is the ultimate verification but is project-specific and can't be generalized into a tool.

osilkin98:
Closing — build-root project build tooling is out of scope for the factory repo (see #491 for the fuller reasoning), and per triage discussion this one appears to have been opened unintentionally.

gx-ai-architect:
# Level 3: Decompilation Analysis — Design and Implementation Plan

## What Level 3 Adds

Levels 1-2 (class inventory + javap API surface) prove **binary compatibility** — the public contract is identical. Level 3 goes deeper: it decompiles the bytecode of every method body and compares the reconstructed Java source between original and rebuilt JARs. This catches **implementation differences** — methods with identical signatures but different internal behavior.

## Tool Choice: CFR 0.152

**CFR** (Class File Reader) is the recommended decompiler:
- Available on Maven Central: `org.benf:cfr:0.152` (2.1MB standalone JAR)
- Single JAR, no dependencies — easy to bundle or download at runtime
- Handles Java 1.5/1.6 bytecode well (important for Spring 3.0.0)
- Produces clean, readable output compared to alternatives
- MIT licensed

Alternatives considered:
- **Procyon:** Good quality but less actively maintained, sometimes fails on old bytecode
- **Fernflower (IntelliJ):** Requires IntelliJ dependency tree, harder to use standalone
- **JD-Core:** GPL licensed, can't decompile some Java 5 patterns correctly

## How Decompilation Comparison Works

```
For each class in both JARs:
  1. Decompile original:  java -jar cfr.jar original.jar --outputdir /tmp/orig/
  2. Decompile rebuilt:   java -jar cfr.jar rebuilt.jar --outputdir /tmp/ours/
  3. Normalize both:      strip comments, normalize whitespace, strip line numbers
  4. Diff:                diff -r /tmp/orig/ /tmp/ours/
  5. Filter noise:        remove known false positives
  6. Report:              any remaining diffs are real implementation differences
```

## The Noise Problem

Without filtering, decompilation comparison is unusable — you get hundreds of false diffs. Here are the known noise sources and how to filter them:

### Noise Source 1: Variable Names
Compilers discard local variable names (unless compiled with `-g` debug flag). The decompiler invents names: `var1`, `var2`, `string`, `n`, etc. Different compiler versions or flags produce different invented names.

**Filter:** Normalize all local variable names to a canonical form (`v0`, `v1`, `v2`...) before diffing.

### Noise Source 2: Constant Pool Ordering  
The constant pool (string literals, class references, method references) can be in any order. Different compilers order them differently. This affects decompiler output ordering of imports and string references.

**Filter:** Sort import statements. Ignore ordering differences in string literal placement.

### Noise Source 3: Control Flow Reconstruction
The same `if/else` chain can be decompiled as:
```java
// Compiler A's bytecode → decompiler output:
if (x > 0) { return true; }
return false;

// Compiler B's bytecode → decompiler output:  
return x > 0;
```
Semantically identical, textually different.

**Filter:** This is hard to filter automatically. Flag these as "structural equivalent — manual review recommended" rather than suppressing them.

### Noise Source 4: Synthetic Methods
Compilers generate synthetic bridge methods, access methods for inner classes, and lambda desugaring. Names like `access$000`, `lambda$0`, `ajc$if_0` vary between compiler versions.

**Filter:** Ignore methods with `ACC_SYNTHETIC` or `ACC_BRIDGE` flags. We already saw this with AspectJ's `ajc$if_0` vs `ajc$if$6f1`.

### Noise Source 5: Debug Info
Line number tables and local variable tables differ between debug/release builds and between compilers. These affect decompiler output (it uses line numbers to place code).

**Filter:** Strip `// line X` comments from decompiler output. Decompile with `--removeinnerclasssynthetics true` flag.

## What Real Differences Look Like (vs Noise)

**Noise (ignore):**
```diff
- int n = list.size();
+ int n2 = list.size();
```
Just a variable name difference.

**Real difference (flag):**
```diff
- return Collections.unmodifiableList(result);
+ return new ArrayList(result);
```
Same signature, different behavior — one returns an unmodifiable view, the other a mutable copy. This would be caused by a different dependency version on the classpath at compile time, or a source code difference.

**Real difference (flag):**
```diff
- if (logger.isDebugEnabled()) {
-     logger.debug("Processing " + name);
- }
+ logger.debug("Processing " + name);
```
The guard check was removed — different optimization or source change.

## Implementation Architecture

```python
def decompile_compare(original_jar, rebuilt_jar, cfr_jar):
    # 1. Decompile both
    decompile(cfr_jar, original_jar, "/tmp/orig")
    decompile(cfr_jar, rebuilt_jar, "/tmp/ours")
    
    # 2. Normalize
    for f in find_java_files("/tmp/orig"):
        normalize(f)  # strip comments, variable names, line numbers
    for f in find_java_files("/tmp/ours"):
        normalize(f)
    
    # 3. Diff
    diffs = diff_directories("/tmp/orig", "/tmp/ours")
    
    # 4. Classify diffs
    results = []
    for d in diffs:
        if is_variable_name_only(d):
            continue  # noise
        if is_synthetic_method(d):
            continue  # noise
        if is_import_reorder(d):
            continue  # noise
        if is_control_flow_equivalent(d):
            results.append({"type": "structural", "review": "manual", **d})
        else:
            results.append({"type": "implementation", "review": "required", **d})
    
    return results
```

## Expected Results for Spring 3.0.0

Based on our build:
- **Same source code** (unmodified v3.0.0.RELEASE tag)
- **Same compiler** (Eclipse JDT 3.3.0)
- **Same JDK** (Sun 1.6.0_45)
- **Different dependencies** (Maven Central vs EBR)

Expected Level 3 output:
- **Most classes:** zero diffs after noise filtering (same source + same compiler = same bytecode)
- **Some classes:** minor diffs from dependency version differences (EBR may have had slightly different versions of transitive deps on the compilation classpath, affecting constant folding or inlined constants)
- **spring-aspects:** known diffs in AspectJ synthetic methods (already caught at Level 2)

## Limitations Even With Level 3

Full decompilation comparison still cannot verify:
1. **Runtime behavior with different dependency JARs on the classpath** — our rebuilt Spring JAR may call the same method on a dependency, but if the dependency JAR is slightly different (EBR repackage vs Maven Central), the behavior at runtime differs. Decompilation only shows Spring's code, not the dependency's behavior.
2. **Reflection and dynamic dispatch** — code that uses `Class.forName()` or `Method.invoke()` can behave differently based on what's on the classpath at runtime, not compile time.
3. **Native methods** — JNI calls are opaque to decompilation.
4. **Serialization compatibility** — two classes with identical source can produce incompatible serialized forms if compiled with different `serialVersionUID` computation (depends on class layout, which depends on compiler).

PR #594, opened 2026-06-19
fix: benchmark workflow skips all jobs on PR trigger

RobotSail:
## Summary

Fixes the matrix `enabled` field check — all benchmarks were being skipped on `pull_request` events because `matrix.enabled == 'true'` (string comparison) didn't match the boolean `true` produced by the expression.

**Fix:** Replace string comparisons with direct boolean checks: `${{ matrix.enabled }}` / `${{ !matrix.enabled }}`.

Also removes the no-op `if: true` on the job level.

## Test plan

- [ ] Verify benchmarks run on this PR (not skipped)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -56,16 +56,15 @@ jobs:
           - benchmark: terminalbench
             default_instance: 'fix-git'
             enabled: ${{ (github.event_name != 'workflow_dispatch') || inputs.benchmark == 'terminalbench' || inputs.benchmark == 'all' }}
-    if: true
 
     steps:
       - name: Skip if not enabled
-        if: matrix.enabled != 'true'
+        if: ${{ !matrix.enabled }}
         run: echo "Skipping ${{ matrix.benchmark }} (not selected)" && exit 0
 
       - name: Gate
         id: gate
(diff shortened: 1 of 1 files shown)


RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** Core boolean fix is correct — all benchmarks were skipped on PR trigger due to string/boolean mismatch. New report job and concurrency handling are well-implemented. Three shell injection risks in workflow (inputs.instance_id flows through bare ${{ }} expressions at lines 101 and 133) — should use env: variables consistently, as was already done for inputs.timeout. Non-blocking since only exploitable by workflow_dispatch-authorized users.

**Experiment:** #0
**Hypothesis:** Fix benchmark workflow boolean check and add PR/release reporting

### Score Comparison

| Metric | Value |
|--------|-------|
| Before | 0.0000 |
| After | 0.0000 |
| Delta | +0.0000 |
| Threshold | 0.0000 |

### Guard Checks

| Check | Result |
|-------|--------|
| tests | ✅ PASS |
| lint | ❌ PASS(pre-existing) |

---
*Posted by Factory CEO*

osilkin98 (approved):

PR #620, opened 2026-06-21
fix: replace :ro bind mount with docker cp for GCP credentials in ProgramBench

RobotSail:
The credentials file was mounted read-only, so chmod failed silently and the agent user couldn't read it. Now uses docker cp + chmod 644 on a regular file instead.

### DIFF
--- a/benchmarks/run-programbench.sh
+++ b/benchmarks/run-programbench.sh
@@ -128,16 +128,10 @@ log "Step 4: Starting cleanroom container"
 RESULTS_DIR="$(mktemp -d /tmp/programbench-results-XXXXXX)"
 echo "    Results directory: ${RESULTS_DIR}"
 
-GCLOUD_MOUNT_ARGS=()
 GCLOUD_ADC="${GOOGLE_APPLICATION_CREDENTIALS:-${HOME}/.config/gcloud/application_default_credentials.json}"
-if [ -f "${GCLOUD_ADC}" ]; then
-    GCLOUD_MOUNT_ARGS=(-v "${GCLOUD_ADC}:/tmp/gcloud-adc.json:ro")
-    echo "    Mounting gcloud credentials"
-fi
 
 docker run -d --name "${CONTAINER_NAME}" \
     -v "${RESULTS_DIR}:/results" \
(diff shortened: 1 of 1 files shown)


RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** Correct fix: :ro VFS mounts are immutable so chmod always fails silently; docker cp creates a writable-layer file where chmod 644 works. Clean diff, improved error handling, no security concerns.


### Score Comparison

| Metric | Value |
|--------|-------|
| Before | 0.0000 |
| After | 0.0000 |
| Delta | +0.0000 |
| Threshold | 0.0000 |

### Guard Checks

| Check | Result |
|-------|--------|
| eval_immutable | ✅ PASS |
| git_clean | ✅ PASS |
| experiment_branch | ✅ PASS |
| scope | ✅ PASS |

### Precheck Gate

```
factory guard could not run (no .factory/config.json in factory repo). Guards assessed manually — all PASS.
```

### Code Review Notes

- Root cause correct: :ro bind mounts are VFS-immutable, chmod cannot change metadata
- docker cp + chmod 644 on writable-layer file is the right fix
- Placement correct: after agent user creation, before Claude Code smoke test
- Error handling improved: failures now abort under set -euo pipefail instead of being swallowed
- Security equivalent or better: no host file path exposed to container

---
*Posted by Factory CEO*

RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** Fix is correct: :ro bind mount made chmod a silent no-op; docker cp creates a writable file. Better error handling (no || true suppression), proper if-guard, clean scope.

**Experiment:** #0
**Hypothesis:** fix: replace :ro bind mount with docker cp for GCP credentials

### Score Comparison

| Metric | Value |
|--------|-------|
| Before | 0.0000 |
| After | 0.0000 |
| Delta | +0.0000 |
| Threshold | 0.0000 |

### Guard Checks

| Check | Result |
|-------|--------|
| scope | ✅ PASS |
| correctness | ✅ PASS |
| security | ✅ PASS |

---
*Posted by Factory CEO*

osilkin98 (approved):
LGTM — factory review passed.

osilkin98 (approved):
LGTM — factory review passed.

PR #624, opened 2026-06-21
feat: add --solver flag to benchmark CI for factory vs claude-code comparison

RobotSail:
Adds solver selection (factory | claude-code | both) to all benchmark scripts and the GHA workflow. Enables side-by-side comparison of factory CEO vs raw Claude Code on the same tasks.


osilkin98 (approved):
LGTM — factory review passed.

osilkin98 (approved):
LGTM — factory review passed.

PR #629, opened 2026-06-21
fix: pass --extra-instruction-path for claude-code solver in TerminalBench

RobotSail:
Without extra instructions, Harbor's claude-code agent asks for confirmation and stops after 50s.


osilkin98 (approved):
LGTM

PR #640, opened 2026-06-21
fix: YAML syntax error in benchmark workflow (inline Python breaking parser)

RobotSail:
The inline python3 -c with colons broke YAML parsing, preventing workflow_dispatch from working. Replaced with a heredoc temp script.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -229,19 +229,23 @@ jobs:
           REF="${{ github.ref }}"
           PR_NUMBER="${{ inputs.pr_number || github.event.pull_request.number || '' }}"
 
+          cat > /tmp/append_history.py << 'PYEOF'
+          import json, sys, os
+          f = sys.argv[1]
+          with open(f) as fh:
+              data = json.load(fh)
+          data['run_id'] = os.environ['RUN_ID']
+          data['commit'] = os.environ['COMMIT']
+          data['ref'] = os.environ['REF']
+          data['pr_number'] = os.environ.get('PR_NUMBER', '')
+          print(json.dumps(data))
+          PYEOF
+
           if [ -d results ]; then
             for f in results/*.json; do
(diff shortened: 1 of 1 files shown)


RobotSail:
## Reviewer: APPROVE

Confirmed the bug is real — the main branch YAML fails to parse (`yaml.scanner.ScannerError` at the inline Python block), while this PR's version parses cleanly.

### What was checked

| Area | Assessment |
|------|------------|
| **YAML syntax** | Heredoc (`<< 'PYEOF'`) keeps Python code inside a shell string, invisible to the YAML parser. Verified with `yaml.safe_load()` — parses cleanly. |
| **Security** | Old code interpolated shell vars into Python strings (`'$RUN_ID'`) — injection-prone. New code uses `os.environ` — safe. Clear improvement. |
| **Correctness** | Same 4 metadata fields, same `json.dumps` output, same append to `$HISTORY`. `os.environ.get('PR_NUMBER', '')` handles empty case. Functionally equivalent. |
| **CI/CD** | `/tmp/` file in ephemeral GHA VM is fine. Script created once outside the loop — marginally more efficient. |
| **Edge cases** | Empty `PR_NUMBER` handled by `.get()` default. Missing results dir handled by `if [ -d results ]`. `/tmp` file overwrite is safe (`cat >`). |

**Issues found:** 0 critical, 0 important, 0 minor.

All CI checks passing (lint, tests on 3.11/3.12/3.13, codecov). Ship it.

RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** Clean fix: replaces inline Python (which broke YAML parsing due to colons) with heredoc temp script. Also improves security by using os.environ instead of shell interpolation. Functionally equivalent, all tests pass.


---
*Posted by Factory CEO*

osilkin98 (approved):
LGTM

PR #650, opened 2026-06-21
fix: benchmark-data branch creation in report job

RobotSail:
Closes #649

## Changes

- Replaced the two-step "Checkout benchmark-data branch" + "Create benchmark-data branch if needed" approach with a single "Setup benchmark-data branch" step
- The new step uses `git ls-remote` to check if the branch exists, then either clones it or initializes it from scratch
- Fixes the failure where the checkout action left a remote that conflicted with the fallback's `git remote add origin`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -217,28 +217,21 @@ jobs:
 
       - uses: actions/checkout@v4
 
-      - name: Checkout benchmark-data branch
-        id: checkout-data
-        uses: actions/checkout@v4
-        with:
-          ref: benchmark-data
-          path: benchmark-data
-          token: ${{ secrets.GITHUB_TOKEN }}
-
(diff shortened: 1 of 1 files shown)


RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** Clean fix replacing broken two-step checkout+fallback with reliable git ls-remote probe. No critical issues, no security concerns, focused diff.

**Hypothesis:** fix: benchmark-data branch creation in report job

### Score Comparison

| Metric | Value |
|--------|-------|
| Before | 0.0000 |
| After | 0.0000 |
| Delta | +0.0000 |
| Threshold | 0.0000 |

### Guard Checks

| Check | Result |
|-------|--------|
| scope | ✅ PASS |

---
*Posted by Factory CEO*

osilkin98 (approved):
LGTM

PR #652, opened 2026-06-21
fix: debug and fix benchmark data append to benchmark-data branch

RobotSail:
Adds debug logging to identify why results.jsonl stays empty. Likely root cause: actions/checkout wipes the downloaded artifacts directory.


osilkin98 (approved):
LGTM — debug

PR #653, opened 2026-06-21
fix: reorder report job steps so artifacts survive checkout

RobotSail:
## Summary

- Move `actions/checkout@v4` before the artifact download step in the report job so it no longer wipes the downloaded `results/` directory
- Remove debug output added during investigation

## Root cause

The report job downloaded artifacts to `results/`, then `actions/checkout@v4` wiped the entire workspace. The subsequent append step couldn't find `results/` and silently exited.

## Test plan

- [ ] Trigger a benchmark workflow run and verify the report job appends results to `benchmark-data/results.jsonl` successfully

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -208,14 +208,14 @@ jobs:
     if: always()
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


osilkin98 (approved):
LGTM

PR #654, opened 2026-06-21
fix: move artifact download after checkout so results/ survives

RobotSail:
actions/checkout wipes the workspace, deleting the downloaded artifacts. Moving the download step after checkout fixes this.


osilkin98 (approved):
LGTM

PR #658, opened 2026-06-21
fix: dashboard data pipeline — solver field, cost extraction, chart labels

RobotSail:
Fixes 4 issues: adds solver as top-level field in results, extracts cost from Harbor agent logs for TerminalBench, improves chart axis labels, and verifies programbench claude-code matrix entry.


osilkin98 (approved):
LGTM

PR #664, opened 2026-06-21
fix: replace average duration/cost bar charts with line charts over time

RobotSail:
## Summary

- Replace the static 2-bar average duration/cost charts (regression) with line charts plotting trends over time
- Each data point = average across all benchmarks for a given `run_id` on main branch
- Two lines per chart: Factory vs Claude Code
- Duration chart: "Average Duration Over Time (Main Branch)" with time-series x-axis
- Cost chart: "Average Cost Per Run Over Time (Main Branch)" with cumulative total spend text above
- Per-benchmark breakdown bar charts preserved below the trend lines

## Changes

- Added `computeRunAverages()` helper that groups main branch entries by `run_id`, computes per-solver average duration and cost for each run, and returns sorted `{x: Date, y: value}` point arrays
- Replaced `renderDurationSection` average chart: `type: 'bar'` → `type: 'line'` with `x: { type: 'time' }` scale
- Replaced `renderCostSection` average chart: same bar→line conversion, added "Total Spend" summary text
- Only `docs/benchmarks.md` modified

### DIFF
--- a/docs/benchmarks.md
+++ b/docs/benchmarks.md
@@ -287,6 +287,58 @@ function computeSolverAverages(mainResults) {
   return result;
 }
 
+function computeRunAverages(mainResults) {
+  const runsByRunId = {};
+  for (const e of mainResults) {
+    const rid = e.run_id || e.timestamp || '';
+    if (!runsByRunId[rid]) runsByRunId[rid] = [];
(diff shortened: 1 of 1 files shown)


osilkin98 (approved):
LGTM

PR #671, opened 2026-06-21
fix: benchmark-eval check only passes when factory jobs ran

RobotSail:
The check was passing unconditionally. Now it inspects result artifacts — only sets success when factory solver results exist and all completed. Claude-code only or disabled benchmarks leave the check pending.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -418,14 +418,49 @@ jobs:
               });
             }
 
-      - name: Update benchmark status to success
+      - name: Update benchmark status
         if: inputs.pr_number != ''
         uses: actions/github-script@v7
         with:
           script: |
+            const fs = require('fs');
             const prNumber = parseInt('${{ inputs.pr_number }}', 10);
             if (!prNumber || isNaN(prNumber)) return;
 
+            // Check factory results in downloaded artifacts
+            let factoryResults = [];
+            if (fs.existsSync('results')) {
+              for (const file of fs.readdirSync('results')) {
+                if (!file.endsWith('.json')) continue;
+                try {
+                  const data = JSON.parse(fs.readFileSync('results/' + file, 'utf8'));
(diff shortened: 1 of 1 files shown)


RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** Correct fix for unconditional success status — logic is sound, edge cases handled, no security concerns.


### Score Comparison

| Metric | Value |
|--------|-------|
| Before | 0.0000 |
| After | 0.0000 |
| Delta | +0.0000 |
| Threshold | 0.0000 |

### Guard Checks

| Check | Result |
|-------|--------|
| eval_immutable | ✅ PASS |
| scope | ✅ PASS |
| git_clean | ✅ PASS |

### Code Review Notes

- Solver field lookup (data.solver
- 
- data.details?.solver) correctly matches both lib.sh top-level and per-benchmark DETAILS_JSON sources
- Edge cases handled: missing results dir, no JSON files, malformed JSON, no factory results all return early correctly
- Empty catch block is acceptable for skip-on-error intent but a debug log would aid troubleshooting
- No security concerns: pr_number parsed to int, file paths from readdirSync not user input

---
*Posted by Factory CEO*

RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** Fix correctly replaces unconditional success with data-driven status check. Solver field lookup handles both JSON sources, edge cases covered, no security issues.

**Experiment:** #0
**Hypothesis:** benchmark-eval status check only passes when factory jobs ran

### Score Comparison

| Metric | Value |
|--------|-------|
| Before | 0.0000 |
| After | 0.0000 |
| Delta | +0.0000 |
| Threshold | 0.0000 |

### Guard Checks

| Check | Result |
|-------|--------|
| scope | ✅ PASS |
| correctness | ✅ PASS |
| security | ✅ PASS |

---
*Posted by Factory CEO*

osilkin98 (approved):
LGTM

PR #746, opened 2026-06-24
fix: add factory init step to eval-baseline CI

osilkin98:
Fixes the eval-baseline workflow by adding `factory init` before `factory eval`, since `.factory/config.json` is gitignored and not present in CI checkouts.

## Changes
- Added `uv run factory init .` step between dependency installation and eval execution in `.github/workflows/eval-baseline.yml`


osilkin98:
Recreating from RobotSail fork

PR #760, opened 2026-06-24
fix: handle transcript JSONL format in ceo.message emitter

colehurwitz:
## Changes

- Updated `_make_ceo_message_emitter()` in `factory/runners/claude.py` to handle both message formats:
  - **Stream-json format** (headless mode): `{"type": "assistant", "message": "Hello..."}`  — message is a string, used directly
  - **Transcript JSONL format** (interactive mode): `{"type": "assistant", "message": {"content": [{"type": "text", "text": "Hello..."}]}}` — text extracted from content blocks
- Added 4 new test cases for transcript format: single text block, multiple text blocks with non-text interleaved, empty content array, and content with only non-text blocks
- Renamed `test_skips_non_string_message` → `test_skips_non_string_non_dict_message` to reflect the updated behavior (dicts are now handled, but other types like int are still skipped)

### DIFF
--- a/factory/runners/claude.py
+++ b/factory/runners/claude.py
@@ -33,13 +33,24 @@ def _make_ceo_message_emitter(project_path: Path) -> Callable[[bytes], None]:
         if not isinstance(parsed, dict) or parsed.get("type") != "assistant":
             return
         message = parsed.get("message", "")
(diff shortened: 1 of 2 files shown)


colehurwitz:
@ceo-review

PR #806, opened 2026-06-26
Fix Node.js eval: detect vitest, auto-detect language for observability

lukeinglis:
## Summary

- **NodeEvaluator detects vitest vs jest** from `package.json` (devDependencies and scripts.test). Projects with vitest run `npx vitest run --coverage` instead of hardcoded `npx jest`. Coverage parsing handles both Jest text-summary and Vitest table formats.
- **Observability auto-detects language** instead of hardcoding `"python"`. TypeScript projects now score correctly (was always 0.0 because it looked for `.py` files).

## Impact

Tested on a Next.js/TypeScript project (Slipstream):
| Dimension | Before | After |
|-----------|--------|-------|
| tests | 0.0 | 1.0 |
| coverage | 0.5 (default) | 0.32 (real) |
| observability | 0.0 | 0.357 |
| composite | 0.458 | 0.612 |
| passed | false | true |

## Files changed

- `factory/eval/languages/node.py`: Added `_detect_test_runner()`, split test execution into jest/vitest paths, extracted `_parse_test_output()` shared parser
- `factory/eval/growth.py`: Changed `"python"` to `"unknown"` on line 167 to trigger language auto-detection

### DIFF
--- a/factory/eval/growth.py
+++ b/factory/eval/growth.py
@@ -164,7 +164,7 @@ def eval_observability(project_path: Path) -> dict:
     try:
         from factory.study import _analyze_observability
 
-        result = _analyze_observability(project_path, "python")
+        result = _analyze_observability(project_path, "unknown")
         score = result.get("observability_score", 0.0)
         fn_cov = result.get("function_coverage", 0.0)
         structured = result.get("structured_logging", False)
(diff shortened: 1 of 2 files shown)


osilkin98:
@ceo-review

PR #880, opened 2026-06-30
fix: use OTel span attributes for Langfuse trace naming

RobotSail:
## Summary
- Replaced raw `urllib` calls to the legacy `/api/public/ingestion` Langfuse endpoint with OTel span attributes (`langfuse.trace.name` / `langfuse.trace.input`) for setting trace names
- The legacy endpoint returns 403 on some self-hosted Langfuse instances while the SDK's OTLP export path works fine — this was causing noisy error logs on every trace
- Removed the `flush()` sleep/retry loop and periodic re-assertion in `TranscriptTailer` since trace naming is now baked into span data at creation time

## Test plan
- [x] All 15 telemetry tests pass
- [x] `ruff check` passes
- [x] Manually verified traces appear with correct names in Langfuse

### DIFF
--- a/factory/telemetry.py
+++ b/factory/telemetry.py
@@ -24,7 +24,6 @@ except ImportError:
 
 _client: object | None = None
 _observations: dict[str, Any] = {}
-_trace_names: dict[str, tuple[str, Any]] = {}  # trace_id -> (name, input)
 
 def is_enabled() -> bool:
     """Check if Langfuse is configured and lazily initialise the client."""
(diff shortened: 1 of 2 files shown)


RobotSail:
## ✅ Factory Review: KEEP

**Verdict:** KEEP
**Reason:** 


### Score Comparison

| Metric | Value |
|--------|-------|
| Before | 0.6712 |
| After | 0.6712 |
| Delta | +0.0000 |
| Threshold | 0.8000 |

---
*Posted by Factory CEO*

osilkin98 (approved):

PR #881, opened 2026-06-30
feat: make langfuse a core dependency

gx-ai-architect:
## Summary

- Move `langfuse>=3.0` from `[telemetry]` optional extra to core dependencies
- Remove try/except import guard in `factory/telemetry.py` — langfuse is now always importable
- Keep `[telemetry]` extra for backward compat (commented as such)

Langfuse is lightweight (~4 MB package + ~1 MB otel deps), and most of its dependencies (`pydantic`, `httpx`, `wrapt`, `packaging`) are already in the dependency tree. Making it a core dep ensures tracing works by default — previously, agents would silently produce no traces when the extra wasn't installed.

## Test plan
- [x] `ruff check factory/telemetry.py` — clean
- [x] `pytest tests/ -x` — 1525 passed (1 pre-existing failure in `verification.md`)
- [x] Verified all 3 deep-qa agents produce Langfuse traces with observations

### DIFF
--- a/factory/telemetry.py
+++ b/factory/telemetry.py
@@ -11,16 +11,12 @@ from pathlib import Path
 from typing import Any
 
 import structlog
+from langfuse import Langfuse
+from langfuse.types import TraceContext
 
 log = structlog.get_logger()
 
-try:
-    from langfuse import Langfuse  # type: ignore[import-not-found]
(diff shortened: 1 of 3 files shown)


gx-ai-architect:
@ceo-review

PR #898, opened 2026-07-01
fix: make langfuse import optional to prevent OTel connection errors

### LINKED ISSUE #884: CEO message streaming fails when langfuse not installed
## Summary

The `TranscriptTailer` (which emits `ceo.message` events to `.factory/events.jsonl`) fails to start when `langfuse` is not installed, even though langfuse is only needed for telemetry/tracing and not for the message streaming functionality.

## Root Cause

In `factory/telemetry.py`, langfuse is imported at module level:

```python
from langfuse import Langfuse
from langfuse.types import TraceContext
```

When `_start_ceo_tailer()` in `cli.py` tries to import `TranscriptTailer`:

```python
from factory.telemetry import TranscriptTailer, begin_span, flush, is_enabled
```

The entire import fails if langfuse isn't installed. The exception is caught and silently returns `None`, so no tailer runs and no `ceo.message` events are emitted.

## Impact

- CEO messages don't stream to `events.jsonl`
- UIs that consume the event stream (like refactory-frontend) don't see planning output
- The failure is silent — no warning unless you add explicit logging

## Fix

Make the langfuse import optional:

```python
try:
    from langfuse import Langfuse
    from langfuse.types import TraceContext
    _HAS_LANGFUSE = True
except ImportError:
    Langfuse = None
    TraceContext = None
    _HAS_LANGFUSE = False
```

This allows `TranscriptTailer` to work without langfuse while preserving telemetry when it is installed.

## Environment

Discovered when running Factory via uv tools install (`uv tool install -e .`) which doesn't include the optional langfuse dependency.

gx-ai-architect:
## Summary
- Wraps `langfuse` imports in `try/except ImportError` so the SDK's OTel exporter doesn't initialize eagerly on module import
- Prevents `localhost:3200` connection refused errors when no Langfuse server is running
- All telemetry functions gracefully no-op when langfuse is unavailable

Closes #884

## Test plan
- [x] `tests/test_telemetry.py` — all 15 tests pass
- [x] `TranscriptTailer` imports successfully without langfuse configured

### DIFF
--- a/factory/telemetry.py
+++ b/factory/telemetry.py
@@ -11,12 +11,17 @@ from pathlib import Path
 from typing import Any
 
 import structlog
-from langfuse import Langfuse
-from langfuse.types import TraceContext
 
 log = structlog.get_logger()
 
-_HAS_LANGFUSE = True
+try:
+    from langfuse import Langfuse
+    from langfuse.types import TraceContext
+    _HAS_LANGFUSE = True
+except ImportError:
+    Langfuse = None  # type: ignore[assignment,misc]
+    TraceContext = None  # type: ignore[assignment,misc]
+    _HAS_LANGFUSE = False
 
 _client: object | None = None
 _observations: dict[str, Any] = {}


gx-ai-architect:
@ceo-review

PR #914, opened 2026-07-01
fix: reorder benchmark CI steps so failure analysis is included in artifacts

RobotSail:
The Analyze failures step was running after Upload results, so analysis files were missing from artifacts and PR comments. Swaps the step order.

Correct order is now:
1. Print summary
2. Analyze failures
3. Upload results (captures analysis markdown in the artifact)


osilkin98 (dismissed):
LGTM — fixes the step ordering so analysis files are captured in artifacts.

RobotSail:
@ceo-review

PR #786, opened 2026-06-25
feat: GRAPH-SPEC behavioral specification engine

mihirathale98:
## Summary

Adds a complete specification engine for generating, validating, updating, and querying behavioral project specifications (`GRAPH-SPEC.md`). Replaces the old `SPEC.md` system (static template from discovery data) with an agent-driven pipeline that produces RFC 2119 normative specifications with 16 mandatory sections.

### New module: `factory/spec/`

- **`generate.py`** — Collects source files, batches by token limit (160K), orchestrates Opus extraction → Researcher annotation pipeline. Produces `GRAPH-SPEC.md` at project root.
- **`parser.py`** — Regex-based Markdown parser that produces a structured `RepoSpec` dataclass (modules, dependency edges, shared contracts, entry points, behavioral sections).
- **`validate.py`** — Three-tier validation: path existence checks, Haiku agent import cross-referencing, and behavioral section completeness verification.
- **`update.py`** — Incremental spec patching via diff scoping (`scope_diff`) and Opus patcher agent. Maps git changes to affected modules for targeted updates.
- **`impact.py`** — Module-centric subgraph extraction for agent context injection (`factory spec impact <module>`).

### New CLI subcommands

- `factory spec generate <path>` — full extraction + annotation pipeline
- `factory spec validate <path>` — run 3-tier validation, output report
- `factory spec scope <path>` — scope git diff against spec modules
- `factory spec update <path>` — patch spec from recent changes
- `factory spec impact <module> --project <path>` — extract module subgraph

### New workflow definitions

- **W₁₁ `spec-generate`** — `extract → gate → annotate → gate → validate → gate` with 16-section completeness enforcement
- **W₁₀ `spec-update`** — `diff_scope → patch → gate → revalidate → gate` for incremental patching
- **`spec_update` node** added to `improve_workflow` and `research_workflow` — conditional, non-blocking, fires after experiment verdict if `GRAPH-SPEC.md` exists

### Agent prompt integration

- **Strategist** — reads GRAPH-SPEC for blast radius assessment, includes `GRAPH-SPEC Diff` section in build plans
- **Builder** — reads GRAPH-SPEC to identify contract consumers when modifying shared types
- **QA** — reads GRAPH-SPEC for targeted verification of change impact predictions
- **CEO** — reads GRAPH-SPEC for experiment triage, uses `factory spec impact` for context injection

### New agent prompts

- `spec_extractor.md` — Opus-powered source analysis → `spec_raw.md`
[...]

PR #932, opened 2026-07-02
test: cover remaining uncovered paths in factory/spec/

mihirathale98:
## Changes
- Added test for oversized file batching in `group_into_batches` (generate.py:142-154): verifies that a file exceeding `char_limit` flushes the current batch, gets its own batch, and subsequent files batch normally
- Added test for `git diff --root` failure path in `_get_diff_text` (update.py:66): verifies `RuntimeError` is raised when the root-diff subprocess returns non-zero

Both `factory/spec/generate.py` and `factory/spec/update.py` now have 100% line coverage.


mihirathale98:
Coverage tests pushed directly to PR #786.

PR #933, opened 2026-07-02
fix: halve spec generation batch token limit

mihirathale98:
## Changes

- Lower `BATCH_TOKEN_LIMIT` in `factory/spec/generate.py` from 160,000 to 80,000 tokens
- The previous limit created batches of ~640K chars which, combined with agent overhead (~50-80K tokens of system prompt, CLAUDE.md, playbook, and task instructions), exceeded the context window causing "Prompt is too long" errors
- New limit produces ~320K char batches, leaving sufficient headroom for agent overhead

## Verification

- All 45 tests in `tests/test_spec_generate.py` pass
- `ruff check` clean


mihirathale98:
Fix pushed directly to PR #786.

PR #934, opened 2026-07-02
fix: parallelize spec extraction across batches

mihirathale98:
## Changes

- Extracted `_extract_batch()` async helper that sends one batch of source files to a single `invoke_agent` call
- Changed `generate_spec()` to dispatch all batches in parallel via `asyncio.gather` instead of stuffing all batch info into one prompt
- `generate_spec()` now writes `spec_raw.md` itself by concatenating batch results, rather than relying on the agent to write it
- Annotation phase unchanged — still runs once on the combined `spec_raw.md`
- Updated tests: added `TestExtractBatch` class, added `test_parallel_batches_concatenated`, removed obsolete `test_missing_spec_raw_raises` (no longer applicable since `generate_spec` writes the file)


mihirathale98:
Pushed to PR #786.

PR #957, opened 2026-07-03
fix: restore Langfuse trace analysis for benchmark runs

### LINKED ISSUE #954: Trace analysis is reporting no trace available for summary
In this run: https://github.com/akashgit/remote-factory/pull/953#issuecomment-4871791496

The eval completed after 10 mins (suspiciously short) and the analysis reports that no trace is available for summary. But we've already seen previous PRs _have_ had successful trace analysis  as you can see here: https://github.com/akashgit/remote-factory/pull/940#issuecomment-4871120299 

Put every PR merged since #940 has seen this failure of not being able to find a matching langfuse trace.

osilkin98:
Closes #954

## Changes

- **Fix env var forwarding (root cause):** `ProgramBenchFactoryCeo` in `factory_harbor_agent.py` forwarded zero Langfuse env vars, causing `telemetry.is_enabled()` to return `False` and no traces to be created. `FactoryCeo` was also missing `LANGFUSE_BASE_URL`. Extracted a shared `COMMON_ENV_VARS` tuple used by both classes to prevent future divergence.

- **Direct trace ID propagation:** `begin_cycle_session()` in `runner.py` now writes the trace ID to `.factory/trace_id.txt`. All 5 benchmark run scripts (`run-swebench.sh`, `run-programbench.sh`, `run-featurebench.sh`, `run-terminalbench.sh`, `run-legacybench.sh`) extract this file after the Harbor run and include `trace_id` in `DETAILS_JSON`. `analyze_failure.py` checks for this direct trace ID first, falling back to time-window search only when absent.

- **Solver-aware analysis:** `analyze_failure.py` now skips trace lookup entirely for `claude-code` solver runs (which never produce factory-managed Langfuse traces), eliminating misleading "No trace found" warnings.

- **Verbose logging:** `find_matching_trace()` now logs the search time window, raw trace count, filtered candidate count, and selected trace ID to stderr when `--verbose` is used.

### DIFF
--- a/benchmarks/factory_harbor_agent.py
+++ b/benchmarks/factory_harbor_agent.py
@@ -9,6 +9,26 @@ from harbor.environments.base import BaseEnvironment
 from harbor.models.agent.context import AgentContext
 
 
+COMMON_ENV_VARS = (
+    "ANTHROPIC_BASE_URL",
+    "ANTHROPIC_MODEL",
+    "CLAUDE_CODE_USE_VERTEX",
+    "ANTHROPIC_VERTEX_PROJECT_ID",
+    "CLOUD_ML_REGION",
+    "GOOGLE_APPLICATION_CREDENTIALS",
+    "CLAUDE_CODE_SUBAGENT_MODEL",
+    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS",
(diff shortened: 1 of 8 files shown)


osilkin98:
@ceo-review

osilkin98:
@ceo-review

PR #968, opened 2026-07-06
fix: re-add auto_merge to swebench workflow using git branch -f

osilkin98:
## Summary

Re-adds `auto_merge` FnNode to the swebench workflow. Without it, the Builder's fix stays on the worktree branch which the factory deletes on exit — Harbor can't find the changes and the benchmark fails despite a correct solution.

Uses `git branch -f $BASE HEAD` instead of `git checkout && git merge` — updates main's ref without checking it out, which works inside worktrees.

## Evidence

[Benchmark run 28759321348](https://github.com/akashgit/remote-factory/actions/runs/28759321348): Builder made the correct `Printable.__slots__` fix in 95s ([trace](https://langfuse-ci.ai-innovation.team/trace/08f5568ce347a3d0a59b3a14e0263c45)), but `passed: 0` because changes never landed on main.

## Test plan

- [x] 161 tests pass
- [x] `factory workflow validate swebench` → VALID (4 nodes, 4 edges)
- [x] E2E clean-room test confirmed `git branch -f` fast-forwards main from inside worktree

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #969, opened 2026-07-06
fix: switch swebench to deterministic executor + fix trace_id extraction

osilkin98:
## Summary

PR #968 was merged with the old `factory ceo --mode swebench` approach which doesn't work — the CEO paraphrases shell commands instead of running them verbatim, causing the auto_merge to update the wrong branch.

This PR fixes three things:

1. **Switch to `factory workflow run swebench .`** — the deterministic executor runs the exact shell commands from the workflow definition, no LLM interpretation
2. **Fix auto_merge** — uses `git update-ref` (bypasses checkout protection) + `git rev-parse --git-common-dir` (detects actual parent branch, not hardcoded `main`) + copies changed files to parent working tree
3. **Fix trace_id extraction** — copies trace_id.txt to `/logs/agent/` (which Harbor exports), benchmark scripts search with `-name` instead of `-path '*.factory/*'`

## Evidence

Verified locally — **5/5 consecutive passes** with `factory workflow run`:
- Reward: 1.0 every run
- Duration: 6-10 minutes
- trace_id captured: `da6a64b8baba596f066551917ff36da9`

Previous approach (`factory ceo --mode swebench`) failed every run because the CEO wrote its own version of the auto_merge command instead of using the SKILL.md's exact text.

## Test plan

- [x] 24/24 swebench workflow tests pass
- [x] `factory workflow validate swebench` → VALID (4 nodes, 4 edges)
- [x] 5/5 local Harbor swebench runs RESOLVED
- [x] trace_id appears in result JSON

### DIFF
--- a/benchmarks/factory_harbor_agent.py
+++ b/benchmarks/factory_harbor_agent.py
@@ -156,8 +156,8 @@ class ProgramBenchFactoryCeo(BaseInstalledAgent):
         await self.exec_as_agent(
             environment,
             command=(
-                "mkdir -p /logs/.factory && "
-                "cp .factory/trace_id.txt /logs/.factory/trace_id.txt 2>/dev/null; "
+                "cp /testbed/.factory/trace_id.txt /logs/agent/trace_id.txt 2>/dev/null || "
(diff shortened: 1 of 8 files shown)


osilkin98:
@ceo-review

PR #979, opened 2026-07-07
fix: rewrite legacybench as lean pipeline — remove broken improve_workflow dependency

### LINKED ISSUE #954: Trace analysis is reporting no trace available for summary

osilkin98:
Closes #954

## Changes

- Rewrote `factory/workflow/contributed/legacybench.py` as a lean 4-node pipeline (`study → builder → gate_verify → auto_merge`) matching the terminalbench/featurebench pattern, instead of composing from `improve_workflow()` which requires `.factory/` infrastructure that doesn't exist in Harbor containers
- Study node scans for legacy source files (*.c, *.f, *.cob, *.java, *.s, *.asm, *.py), shows git status/log, Makefile, build scripts, test files, and attempts output format analysis
- Builder node preserves legacy-specific guidance: preserve original language standard, don't modernize, fix only the specific bug, implement general algorithms (no hardcoding for hidden tests), verify on 3+ inputs
- Gate verify and auto_merge nodes follow the same proven pattern as terminalbench
- Added `legacybench` to `QA_EXEMPT_WORKFLOWS` in both `tests/test_workflow_definitions.py` and `tests/test_skill_export.py` since it no longer has a QA agent node
- Removed dependency on `improve_workflow` and its associated imports

### DIFF
--- a/factory/workflow/contributed/legacybench.py
+++ b/factory/workflow/contributed/legacybench.py
@@ -1,125 +1,190 @@
-"""Legacy-Bench benchmark workflow — evaluates factory against Legacy-Bench.
+"""Legacy-Bench benchmark workflow — lean pipeline for legacy code bugs.
 
-Composes from improve_workflow() with:
-- Researcher: output format analysis guidance
-- Builder: legacy code preservation + hidden test awareness
-- gate_build: legacy preservation enforcement (RELOOP if modernized)
-- gate_qa: output format + decimal arithmetic enforcement
-- auto_merge FnNode for containerized benchmark evaluation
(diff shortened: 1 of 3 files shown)


osilkin98:
@ceo-review

PR #982, opened 2026-07-07
fix: set --max-turns 1000 for headless Claude Code invocations

osilkin98:
## Summary
- Sets `--max-turns 1000` in the Claude runner's `build_command()` for `-p` mode invocations
- Without this, fresh CI installs use Claude Code's default max_turns (as low as 3), causing agents to be killed before finishing work
- Observed in terminalbench CI run where the builder hit `Error: Reached max turns (3)` after only 123s

## Test plan
- [x] Verified `--max-turns` flag is accepted by Claude Code CLI
- [x] `pytest tests/` passes

### DIFF
--- a/factory/runners/claude.py
+++ b/factory/runners/claude.py
@@ -105,6 +105,7 @@ class ClaudeRunner:
             "-p", request.task,
             "--output-format", "stream-json",
             "--verbose",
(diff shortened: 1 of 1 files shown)


osilkin98:
@ceo-review

PR #983, opened 2026-07-07
fix: use vars instead of secrets for LANGFUSE_BENCH_HOST to prevent trace link masking

osilkin98:
## Summary

- Moves `LANGFUSE_BENCH_HOST` from `secrets` to `vars` in the benchmark workflow (5 references)
- GitHub Actions masks all secret values in step summaries and logs, which turns Langfuse trace URLs like `https://langfuse-host/trace/<id>` into `***/trace/<id>` — a relative path that the browser resolves against the GitHub Actions page URL, producing broken links
- The host URL is not a credential (unlike the public/secret keys, which remain as secrets) and doesn't need masking

## Required repo change

Add `LANGFUSE_BENCH_HOST` as a **repository variable** (Settings > Secrets and variables > Actions > Variables tab) with the same value currently in the secret. The secret can be removed afterward.

## Test plan

- [ ] Add `LANGFUSE_BENCH_HOST` as a repo variable with the Langfuse host URL
- [ ] Trigger a benchmark run and verify trace links in the step summary point to the Langfuse host, not GitHub

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -185,8 +185,8 @@ jobs:
           CLAUDE_CODE_DISABLE_ADAPTIVE_THINKING: "1"
           MAX_THINKING_TOKENS: "128000"
           CLAUDE_CODE_EFFORT_LEVEL: "XHIGH"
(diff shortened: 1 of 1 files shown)


osilkin98:
@ceo-review

PR #1000, opened 2026-07-14
fix: increase benchmark timeouts to 2h and forward node timeout in executor

osilkin98:
## Summary

- Increase `AgentNode.timeout` from 1200s to 7200s (2 hours) in all 5 benchmark workflows (featurebench, swebench, terminalbench, legacybench, programbench)
- Forward `node.timeout` from workflow definitions to `invoke_agent` in the executor — previously the node timeout was silently dropped and the hardcoded default of 600s always won
- Bump `max_timeout` wall-clock backstop from 3600s to 14400s (4 hours) so it doesn't kill agents before their configured timeout expires
- Remove `--max-turns 1000` from Claude Code CLI invocation, letting Claude Code use its own default

## Test plan

- [x] All 134 workflow definition tests pass
- [x] All 139 benchmark-specific tests pass (timeout assertions updated)
- [x] Runner tests unaffected (3 pre-existing failures unrelated to these changes)
- [ ] Run featurebench CI to verify agents no longer hit premature timeout

### DIFF
--- a/factory/runners/_subprocess.py
+++ b/factory/runners/_subprocess.py
@@ -38,7 +38,7 @@ async def run_subprocess(
     runner_name: str,
     role: str,
     sanitize: bool = False,
-    max_timeout: float = 3600.0,
+    max_timeout: float = 14400.0,
     on_line: Callable[[bytes], None] | None = None,
 ) -> AgentRunResult:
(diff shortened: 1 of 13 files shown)


RobotSail:
@ceo-review

PR #1002, opened 2026-07-14
fix: convert disallowed-tools tests to async to fix IndexError

osilkin98:
## Changes
- Converted `test_background_command_includes_disallowed_tools` and `test_tmux_command_includes_disallowed_tools` from sync to async test functions
- Removed `try/except Exception: pass` wrappers that silently swallowed failures
- Replaced `asyncio.get_event_loop().run_until_complete()` with direct `await` calls (pytest-asyncio auto mode handles async tests)
- Added `asyncio.sleep` mock in the background test to avoid the 5-second polling delay

Both tests were failing with `IndexError: list index out of range` because the sync `run_until_complete()` call fails under pytest-asyncio's already-running event loop, the exception is caught and ignored, and the subprocess mocks are never called — leaving `call_args_list` empty.

### DIFF
--- a/tests/test_runners.py
+++ b/tests/test_runners.py
@@ -1898,28 +1898,26 @@ class TestDisallowedAgentTool:
                 dt_idx = all_args.index("--disallowedTools")
                 assert all_args[dt_idx + 1] == "Agent"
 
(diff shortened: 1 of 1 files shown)


RobotSail:
@ceo-review

PR #1005, opened 2026-07-15
fix: resolve Langfuse trace misidentification in CI concurrent benchmarks

### LINKED ISSUE #1004: BUG: langfuse traces in CI keep getting misidentified between different runs
There's a bug that's currently making it difficult to identify which trace from langfuse belongs to which agent run. For example, if you see (https://github.com/akashgit/remote-factory/actions/runs/29446644437), the `FeatureBench` run appears to have failed but the error analysis script retrieves the trace belonging to the `TerminalBench` run instead of `FeatureBench`, which seems incorrect.

osilkin98:
Closes #1004

## Changes

- **benchmarks/run-harbor.sh**: Add `FACTORY_BENCHMARK` and `FACTORY_INSTANCE_ID` to the `COMMON_AE` array so benchmark metadata reaches the agent container
- **benchmarks/factory_harbor_agent.py**: Add both env vars to `COMMON_ENV_VARS` tuple so the agent forwards them to the factory subprocess
- **factory/telemetry.py**: Read `FACTORY_BENCHMARK` and `FACTORY_INSTANCE_ID` from `os.environ` and include them in the Langfuse trace metadata dict (only when set)
- **scripts/langfuse/langfuse_client.py**: Add optional `tags` parameter to `list_traces()` for future tag-based filtering
- **scripts/langfuse/analyze_failure.py**: Rewrite `find_matching_trace()` to:
  - First try metadata-based filtering (benchmark + instance_id fields)
  - Fall back to text search only when no metadata matches exist
  - **Remove** the dangerous `if not candidates: candidates = traces` all-traces fallback (return `None` instead)
  - **Replace** `max(latency)` tiebreaker with `min(startTime)` (earliest timestamp)
- **tests/test_telemetry.py**: Add 8 new tests covering metadata propagation in `begin_trace()` and all `find_matching_trace()` logic changes


RobotSail:
@ceo-review

PR #1007, opened 2026-07-16
fix: harden legacybench gate_verify with independent build verification

### LINKED ISSUE #1006: Harden gate_verify for legacybench: verify correctness instead of trusting builder self-report
## Summary

The `gate_verify` GateNode in `factory/workflow/contributed/legacybench/workflow.py:124-143` only checks that the builder committed something and greps the builder's output for success/failure keywords. It does **not** actually compile or run the code. This means a builder that *claims* success but produces an incorrect fix will pass the gate and proceed to `auto_merge`.

This was observed in PR #1005 ([benchmark comment](https://github.com/akashgit/remote-factory/pull/1005#issuecomment-4985932638)): the builder reported fixing all 5 buddy-allocator bugs, but 0/1 hidden tests passed. Because the builder's output contained success keywords ("fixed", "committed", "All fixes committed"), the gate returned `PROCEED` and the RELOOP retry mechanism was never triggered.

## Current behavior

1. Checks `git diff HEAD~1 --stat` — verifies the builder committed *something*
2. Greps builder output (`.factory/reviews/builder-latest.md`) for pass/fail keywords
3. If pass keywords found → `PROCEED` to `auto_merge`
4. If fail keywords found → `RELOOP` back to builder (max 3 retries)
5. If no keywords → defaults to `PROCEED`

## Problem

- The gate trusts the builder's **self-assessment** rather than independently verifying correctness
- For benchmarks with hidden test inputs (where the builder can't run the real tests), this provides no actual quality assurance
- The default-to-PROCEED on no keywords is also dangerous — ambiguous output should trigger re-verification, not auto-merge

## Proposed hardening

- **Compile and run**: The gate should at minimum attempt `make && make test` (or equivalent) and check the exit code before proceeding
- **Run with known inputs**: Even a simple execution with sample inputs and output comparison would catch gross failures
- **Default to RELOOP on ambiguity**: If no clear pass/fail signal is detected, the gate should retry rather than assume success
- **Structured builder output**: Consider having the builder emit a machine-readable verdict (e.g., JSON with test results) rather than relying on keyword grep of prose

## References

- Benchmark comment with full analysis: https://github.com/akashgit/remote-factory/pull/1005#issuecomment-4985932638
- Gate implementation: `factory/workflow/contributed/legacybench/workflow.py:124-143`

osilkin98:
Closes #1006

## Changes
- **gate_verify evaluator_command** (`workflow.py`): Replaced keyword-grep on builder prose with independent build verification — runs `make` and `make test` with 600s timeouts, checks for missing builder output and Makefile, includes last 50 lines of build/test output in reloop feedback
- **_parse_fn_verdict** (`executor.py`): Extracts text after `reloop:` prefix from gate output and uses it as feedback, so builders get actionable error context (e.g. "compilation failed — error on line 42") instead of hardcoded "fn gate requested reloop"
- **tests** (`test_legacybench_gate.py`): 10 test cases covering all gate paths (no commits, missing builder output, make success/fail, test success/fail, no Makefile, no test target) and executor feedback passthrough


RobotSail:
@ceo-review

PR #1015, opened 2026-07-16
fix: expose granular test failure details in ProgramBench gate_verify

### LINKED ISSUE #1014: enhance: expose details about the amount of tests that failed inside of ProgramBench
For `ProgramBench` to be considered a success, the produced binary has to pass a number of tests which are taken from the original codebase. There are a lot of tests but the eval runner only produces a reward of 1 if all of the unit tests passed, else they return 0 even if a single test is failing. 

To get more visibility into the test suite, there should be some way of exposing the following information:
* Ideally we'd want a list of the failing test cases(s)
* If that is not possible, we at least want to have the number of failing test cases

osilkin98:
Closes #1014

## Changes

- **Enhanced `gate_verify` evaluator command** — replaced the simple todos-only check with a comprehensive pipeline: todos → compilation → multi-tier test execution
  - Preserves existing todos.md check as the primary gate condition (backward compatible)
  - Verifies compilation via `compile.sh` with `timeout 7200`
  - Multi-tier test detection: `make test` → `pytest` → `/workspace/test.sh` → graceful skip
  - Writes structured results (command, exit code, output) to `/workspace/test-results.txt`
  - On failure, RELOOPs referencing the results file (respects 200-char truncation limit)
- **Updated builder prompt** — instructs reading `/workspace/test-results.txt` on RELOOP iterations alongside `todos.md`
- **Updated reviewer prompt** — notes that test results may be available at `/workspace/test-results.txt`
- **Added `writes` for test-results.txt** — gate_verify declares `/workspace/test-results.txt` in its writes set (moved from reads since the gate creates this file)
- **Added 10 new tests** covering compilation check, multi-tier test detection, timeout, test-results.txt references in command/writes/builder/reviewer prompts, and backward compatibility

**Note:** The `notes` field (item #2 in the issue) was skipped because `GateNode` does not have a `notes` field — it's only on `FnNode`. Adding it to `GateNode` would require modifying `factory/workflow/primitives.py`, which is outside the declared scope. This can be addressed in a follow-up PR.

### DIFF
--- a/factory/workflow/contributed/programbench/test_workflow.py
+++ b/factory/workflow/contributed/programbench/test_workflow.py
@@ -139,6 +139,68 @@ class TestProgrambenchWorkflow:
         assert "todos.md" in node.evaluator_command
         assert "## TODO" in node.evaluator_command
 
+    def test_gate_verify_checks_compilation(self) -> None:
+        """Gate verifies compilation via compile.sh."""
+        wf = workflow()
+        node = wf.nodes["gate_verify"]
+        assert isinstance(node, GateNode)
+        assert node.evaluator_command is not None
+        assert "compile.sh" in node.evaluator_command
+
+    def test_gate_verify_multi_tier_test_detection(self) -> None:
+        """Gate probes for tests in priority order: make test, pytest, test.sh."""
+        wf = workflow()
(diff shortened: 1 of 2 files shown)


osilkin98:
@ceo-review

PR #1018, opened 2026-07-17
feat: add Harbor-Index benchmark integration

### LINKED ISSUE #1017: dataset: Add Harbor-Index to benchmarks
Harbor Index is a curated dataset comprised of tasks which humans have verified as solvable, but are extremely difficult for agents to solve. 

https://harbor-index.org/

osilkin98:
Closes #1017

## Changes

- **benchmarks/config.sh**: Add `harborindex` to `benchmark_all_names()` and `benchmark_config()` with dataset `harbor-index/harbor-index-1.0`, exact filter style, and `HarborIndexFactoryCeo` agent class
- **benchmarks/factory_harbor_agent.py**: Add `HarborIndexFactoryCeo` class that inherits from `FactoryCeo` without overriding `_get_factory_command()` — the generic CEO approach handles Harbor-Index's diverse meta-benchmark tasks
- **benchmarks/run.sh**: Add `harborindex` to validation case statement and update usage/error messages
- **.github/workflows/benchmark.yml**: Add `harborindex` to workflow_dispatch choices and add factory + claude-code matrix entries with default instance `bix-filter-chip-variants`

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -15,6 +15,7 @@ on:
           - terminalbench
           - programbench
           - legacybench
+          - harborindex
           - all
       instance_id:
         description: 'Instance ID (leave default for smoke test)'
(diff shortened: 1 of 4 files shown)


osilkin98:
@ceo-review

Issue #965, opened 2026-07-05
Fix trace_id.txt persistence across worktree cleanup


osilkin98:
## Summary

`trace_id.txt` is written to the worktree's `.factory/` directory (e.g., `.factory-worktrees/run-xyz/.factory/trace_id.txt`). When the CLI cleans up the worktree after the CEO exits, `trace_id.txt` is deleted. The benchmark script's `find` then can't locate it, resulting in empty `trace_id` in benchmark results and broken failure analysis (analyzer picks up wrong trace).

## Evidence

From benchmark run 28723244671: swebench-factory result had `"trace_id": ""`, causing the failure analyzer to use the legacybench trace instead and report "wrong project."

## Fix

Write `trace_id.txt` to a location that survives worktree cleanup — either the parent project's `.factory/` directory or a stable path like `/tmp/factory-trace-id.txt`.

## Scope

Shared infrastructure change — touches `factory/agents/runner.py` or `factory/cli/ceo.py` where trace_id is persisted.

lambdabaa:
Closing as implemented. _preserve_telemetry() at factory/worktree.py:263-277 copies trace_id.txt from the worktree's .factory/ to the main project's .factory/ before removal. It is called from remove_worktree() at line 367. The telemetry files list is defined as _TELEMETRY_FILES = ("trace_id.txt",) at line 18.

/cc @osilkin98

PR #1021, opened 2026-07-17
fix: use Harbor artifacts in analyze_failure.py when no Langfuse trace exists

### LINKED ISSUE #1004: BUG: langfuse traces in CI keep getting misidentified between different runs

osilkin98:
Closes #1004, #965

## Changes

- **Removed claude-code solver short-circuit** in `main()` — claude-code results now flow through `generate_report()` instead of returning a static dead-end message. Langfuse credential loading and trace search are still skipped for claude-code (since it genuinely has no traces).
- **Added `_find_trial_log()` helper** — discovers `${TIMESTAMP}-${BENCHMARK}-trial.log` files adjacent to the result JSON, reading up to 50KB (tail-truncated for large logs).
- **Enhanced `generate_report()` no-trace path** — when `trace is None`, builds combined text from trial.log + `details.exception`. If LLM is enabled, passes combined text to `run_llm_analysis()` for diagnosis. If LLM is disabled (or fails), renders artifacts under `#### Harbor Artifacts`. Only falls through to "No matching Langfuse trace found" when genuinely nothing is available.
- **Fixed summary mode** — when no trace exists but trial.log/exception are available, passes them to `run_llm_summary()` instead of returning the static "No trace available" message.
- **Added `result_dir` parameter** to `generate_report()` — passed from `main()` as the result JSON's parent directory so the function knows where to look for sibling artifacts.
- **Added 9 tests** covering: claude-code solver no longer short-circuited, trial.log fallback, LLM analysis with trial.log, no-artifacts preserved behavior, summary mode with trial.log, and `_find_trial_log` edge cases (truncation, missing dir, missing file).
- **Captured Harbor output to log file** (`run-harbor.sh`) — Harbor's stdout/stderr is now tee'd to `${TIMESTAMP}-${BENCHMARK}-${BENCHMARK_SOLVER}-harbor.log`. When trial.log isn't found in the Harbor jobs directory, the harbor log is copied to the trial.log artifact path so `analyze_failure.py` picks it up automatically.
- **Fixed tee exit code swallowing** — uses `PIPESTATUS[0]` to preserve Harbor's actual exit code through the tee pipe.
- **Added pre-execution Langfuse wrapper trace** (`lib.sh` + `run-harbor.sh`) — creates a trace before Harbor starts so a trace_id always exists even when the factory CLI crashes before creating its own. The trace is closed with duration/status in the cleanup trap.

### DIFF
--- a/benchmarks/lib.sh
+++ b/benchmarks/lib.sh
@@ -143,6 +143,122 @@ extract_langfuse_hostname() {
     echo "${hostname}"
 }
 
+# create_langfuse_trace — Create a wrapper Langfuse trace via the REST API.
+# Uses Python3 urllib (stdlib) so no pip install is needed.
+# Echoes the 32-char hex trace ID on success, empty string on failure.
+# Env: LANGFUSE_HOST (or LANGFUSE_BASE_URL), LANGFUSE_PUBLIC_KEY, LANGFUSE_SECRET_KEY
+create_langfuse_trace() {
+    local benchmark="${1:-}"
+    local instance_id="${2:-}"
+    local solver="${3:-}"
+    local git_ref="${4:-}"
+
+    python3 -c "
+import json, os, sys, uuid, urllib.request, urllib.error, base64, time
+
+host = os.environ.get('LANGFUSE_HOST') or os.environ.get('LANGFUSE_BASE_URL', '')
+pub_key = os.environ.get('LANGFUSE_PUBLIC_KEY', '')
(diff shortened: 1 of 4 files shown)


osilkin98:
@ceo-review

PR #1026, opened 2026-07-20
update README installation instructions to be global

harrisonstropkay:
So that `factory` can be called from anywhere, we now instruct users to install factory globally.

### DIFF
--- a/.github/workflows/ceo-review.yml
+++ b/.github/workflows/ceo-review.yml
@@ -57,7 +57,9 @@ jobs:
 
       - name: Install factory
         working-directory: factory-trusted
-        run: uv sync --extra telemetry
+        run: |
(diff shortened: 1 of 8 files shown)


gx-ai-architect:
@ceo-review

gx-ai-architect:
if we installed factory globally, what's the point of still doing `uv run factory`, vs just `factory`?

gx-ai-architect:
would also suggest to make install simple, not to put 4 methods of install; 

Just one defualt install, 

and one dev install

gx-ai-architect:
@ceo-review

harrisonstropkay:
@ceo-review

gx-ai-architect:
@ceo-review

PR #1033, opened 2026-07-22
fix: parallel researcher spawning fails in ForkNode workflows (design mode)

ash-ding:
## Summary

**Problem**: In design mode (and all modes with parallel researchers), the CEO agent reports "parallel research spawn timed out" and falls back to sequential execution.

**Two root causes, both equally critical:**

1. **Separate code blocks break shell backgrounding.** `_fork_to_instruction()` in `factory/workflow/skill_export.py` rendered each parallel agent command in a **separate** fenced code block. The CEO (a Claude Code instance) treats each code block as an independent Bash tool call — each runs in its own shell process. So `&` backgrounds a process in shell A, but `wait` runs in shell B where there are no children to await. The backgrounded agents become orphans and `wait` returns immediately.

2. **Missing timeout guidance causes premature kill.** Even when the CEO manages to put all commands in a single shell (e.g. following the text instructions in `ceo.md`), Claude Code's Bash tool has a **default 120-second timeout**. Researchers are configured with `--timeout 600`, but the Bash tool kills the entire shell after 120s — long before any researcher finishes. Without explicit timeout guidance in the SKILL.md, the CEO has no way to know it should override the default.

Both causes must be fixed together: a single code block without timeout guidance still gets killed at 120s; timeout guidance without a single code block still runs `wait` in an empty shell.

**Fix:**
- Extract `_build_agent_command()` helper from `_agent_to_instruction()`
- Rewrite `_fork_to_instruction()` to render all commands + `wait` in a **single** code block
- Add timeout guidance text (e.g. "use --timeout 650") when any agent's timeout exceeds the 120s Bash default
- Bump `_EXPORT_FORMAT_VERSION` in `skill_cache.py` to invalidate stale cached SKILL.md files

**Before** — 4 separate code blocks, no timeout hint (each becomes its own Bash tool call):

    ```bash
    factory agent researcher --review-tag similar ... &
    ```

    ```bash
    factory agent researcher --review-tag techstack ... &
    ```

    ```bash
    factory agent researcher --review-tag pitfalls ... &
    ```

    ```bash
    wait
    ```

**After** — 1 combined code block + timeout guidance:

    ```bash
    factory agent researcher --review-tag similar ... &
    factory agent researcher --review-tag techstack ... &
    factory agent researcher --review-tag pitfalls ... &
    wait
    echo "All parallel agents complete"
    ```

    > Timeout note: the slowest agent above needs ~600s.
    > Use --timeout 650 on this Bash tool call to avoid the default 120s ceiling.

## Files changed

| File | Change |
|------|--------|
| `factory/workflow/skill_export.py` | Extract `_build_agent_command()`, rewrite `_fork_to_instruction()` for single code block + timeout guidance |
| `factory/skill_cache.py` | Add `_EXPORT_FORMAT_VERSION` to cache checksum to invalidate stale SKILL.md |
| `tests/test_skill_export.py` | Add `test_fork_commands_in_single_code_block` and `test_fork_includes_timeout_guidance` |

## Test plan

- [x] All 52 tests in `test_skill_export.py` pass (including 2 new tests)
- [x] All 238 tests in related files pass (`test_splitter.py`, `test_annotations.py`, `test_workflow_definitions.py`, `test_workflow_qa.py`)
- [x] `ruff check` passes on all changed files
- [ ] Manual: run `factory ceo "test idea" --mode design` and verify 3 researchers spawn in parallel without timeout

### DIFF
--- a/factory/skill_cache.py
+++ b/factory/skill_cache.py
@@ -14,6 +14,9 @@ from factory.workflow.primitives import Workflow
 
 log = structlog.get_logger()
 
+# Bump when export code changes affect SKILL.md output (invalidates cached skills).
+_EXPORT_FORMAT_VERSION = "2"
+
 
 def _sort_recursive(obj: object) -> Any:
     """Recursively sort dicts by key and lists by value for deterministic serialization."""
@@ -35,7 +38,7 @@ def _compute_checksum(workflows: dict[str, Workflow]) -> str:
     """
     payload = {name: wf.model_dump(mode="json") for name, wf in sorted(workflows.items())}
     payload = _sort_recursive(payload)
-    blob = json.dumps(payload, sort_keys=True).encode()
+    blob = json.dumps(payload, sort_keys=True).encode() + _EXPORT_FORMAT_VERSION.encode()
     return hashlib.sha256(blob).hexdigest()[:16]
 
 
--- a/factory/workflow/skill_export.py
+++ b/factory/workflow/skill_export.py
@@ -224,13 +224,13 @@ def _format_edges(edges: list[Edge]) -> str:
 # ── node → instruction converters ──────────────────────────────
 
 
-def _agent_to_instruction(
+def _build_agent_command(
     node: AgentNode,
     workflow: Workflow,
     *,
     is_parallel: bool = False,
 ) -> str:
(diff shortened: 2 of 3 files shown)


osilkin98:
@ceo-review

PR #1061, opened 2026-07-26
fix(spike-sort): use DARTsort asymmetric window for spike recovery

colehurwitz:
Closes #1061

## Changes

- **`_compute_quick_templates()`**: Updated to accept `samples_before`/`samples_after` parameters for DARTsort's asymmetric window (42+79=121 samples), with backward compatibility via `half_width`
- **`recover_low_snr_spikes()`**: Computes templates directly from `sorting_final` data using `_compute_quick_templates()` instead of loading from `templates_refined/` — fixes the template/unit ID mismatch (456 templates vs 524 units) caused by DARTsort relabeling units between pipeline stages
- **`definitions.py`**: Removed `templates_refined/` from recovery node's `reads` since it's no longer used
- **Test**: Removed stale `templates_refined/` assertion from recovery node test

### DIFF
--- a/factory/cli/_helpers.py
+++ b/factory/cli/_helpers.py
@@ -16,10 +16,10 @@ log = structlog.get_logger()
 _WIZARD_INPUT_PATH = Path("~/.factory/wizard_input.md")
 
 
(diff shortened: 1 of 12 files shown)


colehurwitz:
Superseded by merge into feature/spike-sort-workflow. The spike recovery implementation is structurally complete but requires DARTsort-native changes (colehurwitz/ds_ref#12) to actually recover spikes.

PR #1072, opened 2026-07-28
fix: apply SPEC Diff from strategy to SPEC.md in improve workflow

### LINKED ISSUE #1064: spec: strategist SPEC Diff is never applied to SPEC.md
## Problem

The strategist produces a `## SPEC Diff` section in its build plan (`.factory/strategy/current.md`) describing which modules are ADDED, MODIFIED, or REMOVED, with plan-spec traceability (`**Implements:**` fields per phase). But **nothing applies that diff to SPEC.md**.

The spec only gets updated post-build via the non-blocking `spec_update` FnNode, which runs `factory spec update` — a blind diff-scope → patch that re-derives changes from the git diff without referencing the strategist's proposed SPEC Diff.

## Impact

- The strategist's careful spec planning is wasted
- The post-build spec update may produce different changes than what was planned
- Plan-spec traceability (`**Implements:**` fields) has no downstream consumer

## Proposed fix

Wire the strategist's SPEC Diff into the spec update path:

1. After the strategist writes `current.md`, extract the `## SPEC Diff` section
2. Pass it to the spec patcher as additional context alongside the git diff scope
3. The patcher applies the planned changes rather than re-deriving from scratch

Alternatively, apply the SPEC Diff as a pre-build step so the spec is updated *before* implementation, not after — giving the builder an accurate spec to work against.

## Files involved

- `factory/agents/prompts/strategist.md` — produces the SPEC Diff (lines 470-503)
- `factory/spec/ops.py` — `update_spec()` and `scope_diff()` 
- `factory/workflow/definitions.py` — `spec_update` FnNode in improve workflow (line 601)
- `factory/agents/prompts/spec_patcher.md` — spec patch agent prompt

mihirathale98:
Closes #1064

## Changes

- Add `factory/spec/apply_diff.py` with `apply_spec_diff(project_path, strategy_path)` that extracts the `## SPEC Diff` section from the strategist's plan, parses ADDED/MODIFIED/REMOVED module entries, and applies them to `SPEC.md`
- Wire as a `FnNode` (`apply_spec_diff`) between `gate_strategy` and `begin` in both the improve and research workflow graphs
- Add `factory spec apply-diff <path>` CLI subcommand for standalone use
- Add 18 tests covering extraction, application (add/modify/remove), missing SPEC.md creation, no-op when no SPEC Diff section, and workflow graph integration
- Regenerate SKILL.md files via `factory workflow export-skills`

## Test plan

- [x] `pytest tests/test_spec_apply_diff.py` — 18 tests pass
- [x] `pytest tests/test_workflow_definitions.py` — 136 tests pass (graph validation)
- [x] `ruff check factory/` — clean
- [x] `mypy factory/spec/apply_diff.py factory/cli/spec.py` — clean
- [x] `factory workflow export-skills` — regenerated successfully

### DIFF
--- a/factory/cli/__init__.py
+++ b/factory/cli/__init__.py
@@ -111,6 +111,7 @@ from factory.cli.review import (
     cmd_review as cmd_review,
 )
 from factory.cli.spec import (
+    cmd_spec_apply_diff as cmd_spec_apply_diff,
     cmd_spec_generate as cmd_spec_generate,
     cmd_spec_impact as cmd_spec_impact,
     cmd_spec_scope as cmd_spec_scope,
--- a/factory/cli/_main.py
+++ b/factory/cli/_main.py
@@ -705,7 +705,7 @@ def build_parser() -> argparse.ArgumentParser:
         "research (autonomous research optimization), review (on-demand PR review), "
         "qa (QA verification pipeline for PRs), "
         "or create (meta-mode for creating or updating factory modes — "
-        "use --focus \"mode_name: change\" to update an existing mode)",
+        'use --focus "mode_name: change" to update an existing mode)',
     )
     p.add_argument(
         "--focus",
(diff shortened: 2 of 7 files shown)


osilkin98:
@ceo-review

PR #1085, opened 2026-07-31
feat: support launching factory in containers

beatsmonster:
WIP


beatsmonster:
Superseded by #1121, which implements the same `factory contained` command on podman + UBI9 instead of NVIDIA OpenShell.

The OpenShell approach needed a gateway process, certificates and a policy engine to stand up, and it denies every process inside the sandbox a PTY — which rules out tmux, screen and dtach alike, and tmux is what attach/detach is built on. The podman version drops all of that, adds an OpenShift target, and has been verified end to end against a real podman host and a real cluster.

The two cannot both merge: they add the same `factory/cli/contained.py`. Closing this one in favour of #1121.

PR #1086, opened 2026-07-31
feat(worktree): add FACTORY_REMOVE_WORKTREE config for worktree retention

mihirathale98:
Closes #1082

## Changes

- Add `FACTORY_REMOVE_WORKTREE` config option (default `true`) — set to `false` to retain run worktrees after sessions for post-mortem debugging and manual follow-up
- Add `_should_remove_worktree(branch)` helper in `factory/worktree.py` using `user_config.resolve()` with 5-tier precedence
- Guard `remove_worktree()` with retention check after the existing `_has_active_sessions()` guard — emits `worktree.retained` event and prints cleanup instructions to stderr when retention is enabled
- Guard `prune_stale()` to skip `run-*` orphan directories when retention is enabled, while still always cleaning `exp-*` experiment orphans
- Fix existing bug: move `worktree.removed` event emission to after actual removal (previously fired even when removal was skipped by `_has_active_sessions()`)
- Register `FACTORY_REMOVE_WORKTREE` in `user_config.py` env_map for `factory config migrate` and add commented template entry
- Add 8 test cases in `TestWorktreeRetention` class covering default behavior, false/0/no values, experiment branch exemption, event emission, and prune_stale guards

Zero call-site changes — the guard is fully centralized inside `remove_worktree()` and `prune_stale()`.

### DIFF
--- a/factory/user_config.py
+++ b/factory/user_config.py
@@ -37,6 +37,7 @@ _CONFIG_TEMPLATE = """\
 # tmux_persist = false                 # Launch agents in tmux windows
 # bg = false                           # Dispatch agents via claude --bg (agent view)
 # bg_agents = false                    # Background sub-agents only (CEO stays foreground)
+# remove_worktree = true               # Set to false to retain run worktrees after sessions
 
 # [credentials.vertex]
 # FACTORY_RUNNER = "claude"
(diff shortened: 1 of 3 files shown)


mihirathale98:
@ceo-review

shivchander:
@ceo-review

PR #1121, opened 2026-08-06
feat(contained): run the factory in a podman container or a cluster pod

beatsmonster:
Runs any factory command somewhere other than your shell — in a podman container on your machine, or in a pod on an OpenShift cluster.

```bash
factory contained -- ceo ~/code/my-project
```

The run happens against a pinned toolchain rather than whatever your machine has accumulated, and it works on a **copy** of your project, so your working tree is never modified. Everything after `--` is passed inward unchanged, so the host never parses the payload and cannot break when the CLI grows.

**Supersedes #1085**, which implemented the same command on NVIDIA OpenShell. That approach needed a gateway process, certificates and a policy engine, and OpenShell denies every process inside the sandbox a PTY — which rules out tmux, and tmux is what attach/detach is built on. This replaces it with an ordinary podman container on UBI9. The two cannot coexist: they add the same `factory/cli/contained.py`.

## Documentation

**[docs/contained/index.md](docs/contained/index.md)** — the user guide: quick start, choosing a target, every flag by target, and a worked example of each command. Published at `/contained/` in the docs site.

`factory contained --help` covers the same surface briefly. Module docstrings carry the reasoning for anything non-obvious in the implementation.

> **It is not a security sandbox.** It makes a run reproducible and keeps it off your working tree. It does not restrict what the agent's code can do, it is not built for code or tenants you do not trust, and it does not replace reviewing the result. `--division` additionally opens an unauthenticated build endpoint on every network interface for the length of the run. The guide says all of this up front.

## Verified commands

Exercised live against podman 5.7.1 on macOS/arm64 and OpenShift 4.21 on ROSA.

| Command | What it proves | |
|---|---|---|
| `factory contained verify` | Three checks on a virgin machine; the footer names only what `setup` can fix | ✅ |
| `factory contained setup` | Starts a stopped podman machine, pulls the image, ends green; idempotent on re-run | ✅ |
| `factory contained setup` (interactive) | Target prompt; survives a closed stdin | ✅ |
| `factory contained -- backlog-list <proj>` | copy → mount → path rewrite → exec → relay; ten lines of output, name first | ✅ |
| `factory contained -- study <proj>` | A real agent call; observations land in the copy, host tree untouched | ✅ |
| `factory contained ls` | One table; distinguishes "none", "could not look" and "target not set up" | ✅ |
| `factory contained attach <name>` | tmux over a pty; `Ctrl-b d` detaches, the run continues | ✅ |
| `factory contained attach` (after `exit`) | Revives the dead pane into a usable shell | ✅ |
| `factory contained sync <name>` | Branch and merge command; merges nothing | ✅ |
| `factory contained rm <name>` | Prompts while active; discloses the worktree and branch it leaves behind | ✅ |
| `podman stop <name>` | Exit 143 immediately — SIGTERM honoured, factory is not PID 1 | ✅ |
| Broken workspace | Aborts at `assert:git_usable` before any agent call, with a `Try:` line | ✅ |
| `FACTORY_CONTAINED_DRY_RUN=1 …` | create → five assertions → run; provisions nothing | ✅ |
| `--forward` / `--env` | Unset and malformed values rejected before anything is created; secrets redacted | ✅ |
| Wrong-target flags, flags after a subcommand, typos | Rejected at parse time with the corrected command | ✅ |
| `factory contained --division -- …` | Server up, reachable from inside the container, banner states the exposure | ✅ |
| `--division -- agent builder --task "list your tools"` | Agent names the podman tools from its brief | ✅ |
| `--division -- ceo … --focus "add a Containerfile"` | **Full loop**: agent wrote a Containerfile, built `localhost/rta:latest` on the host engine, image runs `--help` | ✅ |
| No `--division` | Nothing started, no `.mcp.json`, nothing on :8430 | ✅ |
| `--target k8s … verify` / `setup` | Manifest printed, applied with your own credentials, then green — including inference reachability probed from inside the cluster | ✅ |
| `--target k8s --namespace X bundle` | Valid namespace-scoped YAML; never grants `pods/exec` | ✅ |
| `--target k8s -- backlog-list <proj>` | pack → PVC → initContainer unpack → rewrite → exec → relay | ✅ |
| Secret scan | gitleaks names `.env:1` and refuses the upload; `--yes` overrides and says so | ✅ |
| Pod deleted, then `sync` | Work survived the pod and came back | ✅ |
| `--target k8s attach` / `rm` | tmux over `oc exec -it`; `rm` leaves the PVC alone | ✅ |
| `--target k8s --division` | Sidecar on its own `oc` image; `start_build` → OpenShift `Build` → validation pod ran the built image | ✅ |
| RBAC boundary | `pods/exec` denied to the ServiceAccount, asserted by SubjectAccessReview | ✅ |

**Not verified:** an agent inside a cluster pod listing its own cluster tools — that needs a credentials Secret in the namespace, which is an operator's decision. Everything else about that surface is verified: the registration is written, `oc` is absent from the image, and the `start_build` server answers a real MCP handshake from inside the pod.

## Known limitation

**The runtime image is not published yet.** `ghcr.io/akashgit/remote-factory/factory-runtime:latest` returns 403, so `setup` currently fails at the pull. The publishing workflow is in this PR (`.github/workflows/runtime-image.yml`) but has never run. Until it does, build locally and set `FACTORY_CONTAINED_IMAGE`; `setup` prints both routes on failure.

## Notes for review

- **One change reaches beyond this feature.** `factory/__init__.py` gains a filtering logger, INFO by default with `FACTORY_LOG_LEVEL=debug` to opt in. Before it, `log.debug` and `log.info` rendered identically, so a routine run buried its own output under internal event names. It affects every command. No test asserts on info-level output, and the suite is unchanged against baseline.
- Everything else is new files under `factory/contained/`, `factory/podman.py`, and four lines of CLI registration.
- The suite has ~62 pre-existing failures in the workflow, spec and tmux suites. They reproduce on `main` and with this branch stashed.


osilkin98:
@beatsmonster looks like the unit tests are failing, could you please take a look?

RobotSail:
@ceo-review

PR #1132, opened 2026-08-07
fix: sanitize subprocess output to prevent Claude Code rendering corruption (#1123)

### LINKED ISSUE #1123: Claude code rendering bug
Image

There's a bug (see attached image) where the factory output stream prints in a way that breaks/overlaps some offset rendered by claude-code. I'm not exactly sure on the sequence of actions or machinery breaking it, but the attached screenshot show  an example.

osilkin98:
Closes #1123

## Changes

- **factory/runners/claude.py** — Added `sanitize=True` to the `run_subprocess()` call in `headless()`, matching the Bob runner's existing pattern. This strips ANSI/VT escape sequences from streamed subprocess output before they reach Claude Code's Ink renderer, preventing render state desynchronization.

- **factory/cli/_helpers.py** — Added an early-return TTY guard at the top of `_show_spinner()`: `if not sys.stderr.isatty(): return`. This prevents the spinner's carriage-return writes from reaching Claude Code's Ink renderer when stderr is a pipe, eliminating cursor position desynchronization.

- **factory/runners/_stream.py** — Extended `strip_ansi()` with a carriage-return stripping post-pass after the existing ANSI regex substitution. `data.replace(b"\r\n", b"\n").replace(b"\r", b"")` preserves Windows-style CRLF as LF while stripping bare CR that would corrupt Ink's cursor tracking from subprocess progress bars.

- **tests/test_runners.py** — Updated `test_strip_ansi_preserves_plaintext_and_newlines` to expect bare CR stripping, and renamed `test_claude_runner_does_not_sanitize` to `test_claude_runner_sanitizes` to match the new behavior.

### DIFF
--- a/factory/cli/_helpers.py
+++ b/factory/cli/_helpers.py
@@ -166,6 +166,8 @@ _BRAILLE_FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇"
 
 def _show_spinner(stop_event: threading.Event) -> None:
     """Braille spinner on stderr. Respects NO_COLOR."""
+    if not sys.stderr.isatty():
+        return
     use_color = not os.environ.get("NO_COLOR") and sys.stderr.isatty()
     idx = 0
     while not stop_event.is_set():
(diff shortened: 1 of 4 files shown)


osilkin98:
@ceo-review

PR #1142, opened 2026-08-07
feat: workflow tool — stale file fix, overview/curr/dry-run, compact next

### LINKED ISSUE #222: Finalize gate overrides KEEP to REVERT on pre-existing smoke test failures

xukai92:
Closes #222

## Changes

- **Fix stale file bug**: `tool_init` records `started_at` timestamp in state.json. `_detect_artifact` checks file mtime against session start, ignoring leftover files from prior runs.
- **New `tool_overview` subcommand**: Renders the full workflow map with ✓/▶/○ completion markers. Does not advance the pointer.
- **New `tool_curr` subcommand**: Shows current node details without advancing or auto-submitting.
- **`--dry-run` for `tool next`**: Runs the auto-submit scan but does not persist state changes or emit events. Preview what next would do.
- **Compact `tool next` output**: Returns only the current node task details (via `_format_node_task`) instead of the full progress map. The progress map lives in `overview` now.
- **Inject overview into system prompt**: `_tool_exec_protocol` now appends the workflow map so the CEO has full visibility at session start.
- **6 new tests**: stale file ignored, fresh file detected, overview shows all nodes, curr shows current, next dry-run, next compact output.

### DIFF
--- a/factory/cli/_ceo_helpers.py
+++ b/factory/cli/_ceo_helpers.py
@@ -65,15 +65,23 @@ def _tool_exec_protocol(wt_path: Path) -> str:
         "\n"
         "You are executing the workflow using factory tool commands instead of "
         "following a SKILL.md playbook.\n"
-        "\n"
-        "## Commands\n"
+    )
+
+    if overview:
+        protocol += (
+            "\n## Workflow Map\n"
+            "\n"
+            f"{overview}\n"
+        )
+
+    protocol += (
+        "\n## Commands\n"
         "\n"
         f"  factory workflow tool next {p}\n"
         f"  factory workflow tool submit {p} --node <NODE_ID> <<'TOOL_OUTPUT'\n"
         "  <your output>\n"
         "  TOOL_OUTPUT\n"
         f"  factory workflow tool status {p}\n"
-        f"  factory workflow tool overview {p}\n"
         f"  factory workflow tool curr {p}\n"
         "\n"
         "## Protocol\n"
@@ -102,13 +110,6 @@ def _tool_exec_protocol(wt_path: Path) -> str:
         "do not write code\n"
         '- Start by running "next" to get your first task\n'
     )
-
-    if overview:
-        protocol += (
-            "\n## Workflow Map\n"
-            "\n"
-            f"{overview}\n"
-        )
 
     return protocol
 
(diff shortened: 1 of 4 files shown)


RobotSail:
@ceo-review

xukai92:
@ceo-review

xukai92:
@ceo-review

xukai92:
@ceo-review

PR #1167, opened 2026-08-11
feat(workflow): add design_workflow_v2 recomposed with SubWorkflowNode

mihirathale98:
Recompose design mode using SubWorkflowNode references instead of inline
nodes. Study and research-pipeline run IN PARALLEL via fork_observe/
join_observe. Build-verify is a single SubWorkflowNode reference.

The new workflow has fewer nodes than the original (~12 vs ~20+) while
preserving identical behavior. Registered as 'design-v2' alongside the
original 'design' workflow. Also registers 'strategize-user' and
'strategize-agent' as separate registry entries.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/factory/workflow/definitions.py
+++ b/factory/workflow/definitions.py
@@ -45,6 +45,7 @@ __all__ = [
     "DOC_FRESHNESS_GATE_PROMPT",
     "build_workflow",
     "design_workflow",
+    "design_workflow_v2",
     "improve_workflow",
     "research_workflow",
     "meta_workflow",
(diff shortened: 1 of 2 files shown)


osilkin98:
@mihirathale98 Looks like this has some conflicts against the target branch. Also, can you double-check the target branch? I was under the impression we were targeting the `v2` branch

osilkin98:
@ceo-review

PR #1001, opened 2026-07-14
fix: design mode routing preserves ceo_mode='design' for ALL design invocations

osilkin98:
Closes #999, addresses #908

## Changes

- **Fix design mode routing (ceo.py:580):** Replace `elif design_existing:` with `elif mode == "design":` so that ALL `--mode design` invocations (both new ideas and existing projects) route to `ceo_mode="design"`. Previously, new ideas via `--mode design` fell through to `ceo_mode="build"`, bypassing the design workflow and its user approval gate.

- **Update design_idea task string:** Remove incorrect "proceed to Build mode" language. The design workflow handles everything end-to-end — after user approval at the strategy gate, the workflow continues to implementation automatically.

- **Update and expand test coverage:** Rename `test_design_idea_routes_to_build` → `test_design_idea_routes_to_design` with assertions that `design_idea` gets "Run design mode" (not "Run Build mode"). Add `test_research_ideation_routes_to_build` verifying that `--mode research` still correctly routes to `ceo_mode="build"`. Fix `TestBuildCeoTaskDesign` tests to pass `mode="design"` instead of `"build"` for design_idea cases.

- **Diagnostic warning in state.py** (from prior commit): When `.factory/` exists but `config.json` is missing, `detect_state()` logs a warning with a hint to run `factory init`.

## Test plan

- [x] All 13 targeted tests pass (`pytest -k 'CeoMode or BuildCeoTaskDesign'`)
- [x] `test_design_idea_routes_to_design` verifies new idea gets design workflow
- [x] `test_research_ideation_routes_to_build` verifies research still gets build mode
- [x] `test_design_existing_routes_to_design` unchanged, still passes

PR #1172, opened 2026-08-11
fix: design mode no longer instructs CEO to transition to Improve mode

### LINKED ISSUE #1171: Design mode should not instruct the CEO to move to improve mode
@mihirathale98 Discovered that the design mode has a prompt which instructs it to move to improve mode for existing projects. We generally do not want anything to use the current improve mode because it has a lot of problems and cannot be trusted how we'd like. PR #1001 is related to this bug, because it fixed another routing bug which uncovered this prompt-level issue that began instructing existing factories to move to improve mode. 

We want the factory to use the correct version of the design mode prompt in the correct context for existing factories, and not move to improve mode. I'm just not sure what it should move to in its place, so the CEO design for this issue should understand the context of how #1001 worked back then when the other modes were not considered "regressive" while for existing projects it was. 

Image

RobotSail:
Closes #1171

## Changes

- **Fix task text** (`factory/cli/_task_builder.py`): Replace stale "transition to Improve mode" language in the `design_existing` branch with "the workflow continues to implementation automatically after approval". The sibling `design_idea` branch was already fixed by PR #1001; this patch closes the gap.
- **Set terminal flag** (`factory/workflow/definitions.py`): Add `wf.terminal = True` to `design_workflow()` (non-just_plan branch), preventing `_chain_modes()` from auto-chaining Improve mode after design completes in headless/`--auto-approve` runs. Both `founder` and `plan` modes already set this flag.
- **Fix CLAUDE.md doc**: Update the design mode description to say "continues to implementation automatically after approval" instead of "transitioning to Improve mode".
- **Update test** (`tests/test_workflow_definitions.py`): Rename `test_design_not_terminal` → `test_design_is_terminal` and assert `terminal is True` to prevent regressions.


RobotSail:
@ceo-review

PR #1202, opened 2026-08-12
fix: add timeout guidance for parallel researchers in workflow skills

### LINKED ISSUE #1201: Design mode: Parallel researcher agents timeout due to Bash tool 120s default
## Problem

When running design mode with parallel researchers, the workflow frequently fails with a timeout error. The issue occurs during Phase 2 (Research - Parallel) when spawning 3 concurrent researcher agents.

### Observed Behavior

**Failure scenario:**
- 3 researcher agents are launched in parallel via shell backgrounding (`&`)
- The Bash tool has a default timeout of 120 seconds (2 minutes)
- All 3 researcher agents are killed when the timeout is reached
- Each researcher typically needs 5-10 minutes to complete web research and write findings

**Recovery behavior varies:**
- **Good case**: Claude Code retries the entire research phase with a longer timeout (e.g., 600 seconds / 10 minutes)
- **Bad case**: The CEO agent attempts to reuse partially-written or stale result files from the previous failed attempt, leading to degraded performance and incomplete research synthesis

### Example Evidence

From actual run logs (`run-1c530a06`), we observed two consecutive Bash tool calls:

**First attempt (failed):**
```
Tool ID: toolu_vrtx_01MJnkFY1XPkB8m9bBNnZvKL
Timeout: None  ← Uses default 120s timeout
Status: Timeout failure after 2 minutes
```

**Second attempt (succeeded):**
```
Tool ID: toolu_vrtx_01TBtMhmYJ4FXmAN18xLvTbS
Timeout: 600000  ← Explicitly set to 10 minutes
Status: Success — all 3 researchers completed
```

### Result Files Affected

When parallel researchers fail, the following files may be incomplete or missing:

- `.factory/reviews/researcher-similar-latest.md`
- `.factory/reviews/researcher-techstack-latest.md`
- `.factory/reviews/researcher-pitfalls-latest.md`
- `.factory/strategy/research-similar.md`
- `.factory/strategy/research-techstack.md`
- `.factory/strategy/research-pitfalls.md`

If the CEO attempts to read these incomplete files, the subsequent strategy synthesis will be based on partial or stale data.

## Root Cause

The issue is caused by the **Bash tool's default 120-second timeout**. When the CEO agent reads `skills/workflow-design/SKILL.md` and executes the parallel research phase, it calls the Bash tool without explicitly setting a timeout parameter. The Bash tool then kills the entire shell process (including all 3 backgrounded researcher subprocesses) after 120 seconds.

**Important note**: The parallel execution mechanism itself works correctly. Claude Code intelligently merges the shell commands into a single Bash tool call, so all 3 `factory agent researcher ... &` commands and the `wait` statement run in the same shell process. The problem is purely the timeout constraint.

## Proposed Fix

Add timeout guidance to `skills/workflow-design/SKILL.md` in the Phase 2 (Research - Parallel) section:

```markdown
## Phase 2: Research (Parallel)

Spawn 3 agents in parallel:

```bash
factory agent researcher --review-tag similar --task "..." &
factory agent researcher --review-tag techstack --task "..." &
factory agent researcher --review-tag pitfalls --task "..." &
wait
echo "All researchers complete"
```

**Important:** Run ALL commands above in a **single** Bash tool call 
with timeout set to at least 600 seconds (10 minutes).
```

This guidance will instruct the CEO agent to set an appropriate timeout when invoking the Bash tool, preventing premature termination of the research phase.

## Related

This issue is related to closed PR #1033, which attempted to fix this problem by merging code blocks. However, investigation shows that Claude Code already merges the blocks automatically — the only necessary fix is adding the timeout guidance.

## Impact

- **Frequency**: Occurs on almost every design mode run that reaches the parallel research phase
- **Severity**: Medium — usually recoverable via retry, but causes context waste and occasionally produces degraded results when stale files are reused
- **User experience**: Confusing for users who see timeout errors followed by unexplained retries or incomplete research synthesis

### PR DESCRIPTION
**Pull request by ash-ding**
Closes #1201

## Changes

- **factory/workflow/skill_export.py** — Modified `_fork_to_instruction()` to emit timeout guidance when parallel agents have timeout > 120s. Calculates max timeout among fork targets and adds a guidance block instructing the CEO to set Bash tool timeout appropriately.
- **tests/test_skill_export.py** — Added `timeout` parameter to `_make_agent()` test helper. Added two new tests: `test_fork_includes_timeout_guidance_when_needed` (verifies guidance appears when max timeout > 120s) and `test_fork_omits_timeout_guidance_when_not_needed` (verifies no guidance when timeout <= 120s).

## Problem

When design mode spawns 3 parallel researcher agents (via ForkNode), they frequently timeout because:
- Each researcher needs 5-10 minutes to complete web research
- The Bash tool has a default 120-second timeout
- Even though each `factory agent researcher --timeout 600` sets 10 minutes, the Bash tool kills the entire shell process after 2 minutes

## Solution

Emit timeout guidance in generated SKILL.md files when parallel agents exceed Bash tool's default timeout. The guidance instructs the CEO agent to set an appropriate timeout parameter when invoking the Bash tool.

Example generated guidance:
```markdown
**Important:** Run ALL commands above in a **single** Bash tool call 
with timeout set to at least 600 seconds.
```

This fix is minimal and surgical — it adds guidance text only when needed (max timeout > 120s) and lets the CEO agent handle the timeout parameter naturally.

## Test plan

- [x] All 54 tests in `tests/test_skill_export.py` pass
- [x] `ruff check` passes clean
- [x] `mypy` type check passes
- [x] New tests verify guidance appears when needed and is omitted when not needed

### DIFF
--- a/factory/workflow/skill_export.py
+++ b/factory/workflow/skill_export.py
@@ -597,6 +597,20 @@ def _fork_to_instruction(node: ForkNode, workflow: Workflow) -> str:
         if isinstance(workflow.nodes.get(tid), AgentNode)
     ]
     if agent_nodes:
+        # Calculate the maximum timeout among all parallel agents
+        max_timeout = max(
+            (node.timeout or 600 for node in agent_nodes),
+            default=600
+        )
+
+        # Add timeout guidance if max_timeout exceeds Bash tool's default (120s)
+        if max_timeout > 120:
+            lines.append("")
+            lines.append(
+                f"\n**Important:** Run ALL commands above in a **single** Bash tool call "
+                f"with timeout set to at least {max_timeout} seconds.\n"
+            )
+
         from factory.workflow.verification import compile_fork_verification
 
         verify_script = compile_fork_verification(agent_nodes)
--- a/tests/test_skill_export.py
+++ b/tests/test_skill_export.py
@@ -39,6 +39,7 @@ def _make_agent(
     prompt: str = "",
     reads: set[str] | None = None,
     writes: set[str] | None = None,
+    timeout: int | None = None,
 ) -> AgentNode:
     return AgentNode(
         id=id,
@@ -47,6 +48,7 @@ def _make_agent(
         prompt_template=prompt,
         reads=reads or set(),
         writes=writes or set(),
+        timeout=timeout,
     )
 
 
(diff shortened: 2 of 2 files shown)


ash-ding:
Closing to recreate from correct base

PR #1159, opened 2026-08-10
Extract research subgraph into modular helper

colehurwitz:
Closes #1155

## Changes

- Added `ResearcherConfig` frozen dataclass and `_research_subgraph()` helper function to `definitions.py`, following the exact `_deep_qa_subgraph()` pattern
- Replaced ~90 lines of inline research node definitions in `build_workflow()` with a `_research_subgraph()` call using build-mode configs (similar/techstack/pitfalls with `post_check_min_size=50`)
- Replaced ~90 lines of inline research node definitions in `create_workflow()` with a `_research_subgraph()` call using create-mode configs (existing/intent/practices with no post checks)
- Created `factory/workflow/research.py` as a standalone `research-standalone` workflow module (mirrors `deep_qa.py` structure)
- Registered `research-standalone` in `_get_builtin_registry()` using the lazy import pattern
- Added 31 tests covering: subgraph helper (node/edge counts, post_checks behavior), workflow preservation (build/create/design produce identical node IDs and edge tuples), and standalone workflow (valid graph, trigger, registry)
- Design mode requires no changes — it calls `build_workflow()` and inherits the refactored research subgraph
- No functional changes — purely structural modularization

PR #1208, opened 2026-08-12
fix(contained): verify/setup UX — progress, fail-fast, flag order, guided credentials, and the CI publish break

beatsmonster:
Five fixes to `factory contained`, found while preparing a real cluster run against ROSA. Each commit stands alone.

## The CI break (do this first)

`0f656616` — **the runtime image builds twice and publishes nothing.** Every `factory contained --target k8s` run dies on `ImagePullBackOff` for `factory-runtime:latest`, and the build logs show two *successful* pushes. The `Export the digest` step writes each arch's digest out as a filename, verbatim (`sha256:37f2…`); `upload-artifact@v4` rejects a path containing a colon, so the step after a good push fails, which fails the `build` job, which skips `manifest` (`needs: build`) — and `manifest` is the only thing that creates a tag. The registry ends up holding ~1.7 GB of layers with `"tags": null`. Confirmed directly: fetch by digest → 200, fetch `:latest` → 404, `tags/list` → `null`. Fixed by stripping the prefix on the way in (`Assemble the manifest list` already puts it back).

Because the workflow only triggers on push to main, **this fix does nothing until it lands there.** Once merged, `:latest` publishes on the merge commit; or run `runtime-image.yml` via `workflow_dispatch` to publish immediately.

## The four UX fixes

`8c2c180e` — **a slow step now says what it is waiting for.** `verify` went silent for up to three minutes during the in-cluster inference probe and was read as a hang. A transient status line (`style.activity()`) draws nothing for 5s — so fast checks are unchanged — then rewrites one line in place with the pod's state and a clock, and erases itself on exit. Gated on `can_rewrite()`, not colour: `FORCE_COLOR` in CI must not fill a log with carriage returns. Degrades to one plain line per change off a TTY; `FACTORY_NO_PROGRESS=1` opts out. Brings `read_secret()` (asterisk-masked) and `select()` for the credentials step.

`7f4dceab` — **runtime flags work on either side of the subcommand.** `factory contained verify --target k8s` was rejected; it is now the same command as `--target k8s verify`. One shared flag table feeds both the real parser and a tail parser built with `SUPPRESS` defaults, so an absent flag never clobbers one typed on the other side. The same flag on both sides with different values is an error, not a silent winner.

`b9cd92be` — **three cluster states the runtime described wrongly** (one commit; interleaved hunks in `k8s_setup.py`):
- *A pod that cannot start is waited on for the full timeout.* `oc wait --timeout=180s` is blind — an unpullable image cost the whole three minutes, then reported "the probe produced no output". `classify_pod`/`poll_pod` read the pod document and fail fast on `ImagePullBackOff` etc. with the kubelet's own message; `ContainerCreating` is deliberately not capped (a first pull is legitimately slow). `wait_for_container` on the run path shared the same 300s defect.
- *An expired login was reported as a namespace full of missing objects.* Every `oc get` failed with `Unauthorized`, every failure was read as `NotFound`, and setup offered to create five objects that already existed. Only `NotFound` now means absent; a new `cluster_login` check does one authenticated round trip and stops early, naming `oc login`.
- *The credentials Secret was left entirely to the user*, so every prepared namespace ended one check short. `setup` now offers to create it — typed/env-var/file sources, backend validation up front. Material never reaches an argv (`oc apply -f -` on stdin, not `--from-literal`), the manifest is JSON not YAML, nothing is echoed or logged but key names and lengths.

## Verification

- `ruff` clean; `mypy` unchanged (same 15 pre-existing errors).
- 656 contained tests pass; full suite adds no new failures attributable to this branch.
- **Heads-up, not mine:** `tests/test_workflow_research.py` has 4 failures that reproduce on clean `origin/main` (from #1159). CI here will show them red until that's fixed upstream.


osilkin98:
@beatsmonster Just needs more patch coverage and we can merge it

osilkin98:
@ceo-review

osilkin98:
@ceo-review

PR #1232, opened 2026-08-13
fix: plugin modes rejected and pre-hooks crash on missing path

shivchander:
## Summary

- **Bug 1**: `_validate_ceo_flags` checked `--mode` against the static `CEO_MODES` list instead of `get_all_ceo_modes()`, which includes plugin-registered modes. Plugin modes like `attack` or `buildroot` were rejected as "unknown mode" even though they were properly registered via `add_modes()`.

- **Bug 2**: `cmd_ceo` ran `assert raw_path is not None` (line 43) before invoking pre-hooks (line 106). Plugin modes that create their own work directory via `add_ceo_pre_hook` — the whole point of the pre-hook API — crashed on the assert because no positional path argument was provided. The pre-hook that would supply the path never got to run.

## Changes

- `_ceo_helpers.py`: Use `get_all_ceo_modes()` instead of `CEO_MODES` for mode validation. Defer the "no path provided" error when pre-hooks are registered (they may supply the path).
- `ceo.py`: Move pre-hook invocation before the path-is-None check. Replace the `assert` with a proper error message that only fires after hooks have had their chance to supply a path.

## Test plan

- [x] Verified plugin-registered modes pass validation (`get_all_ceo_modes()` includes them)
- [x] Verified pre-hooks can supply `raw_path` when none provided on CLI — flow reaches `_execute_ceo`
- [x] Verified no-path + no-hooks case still produces the original error message
- [x] Lint passes (`ruff check`)
- [x] Imports clean (`python -c "import factory.cli.ceo; import factory.cli._ceo_helpers"`)

### DIFF
--- a/factory/cli/_ceo_helpers.py
+++ b/factory/cli/_ceo_helpers.py
@@ -233,7 +233,7 @@ def _validate_ceo_flags(
         if focus:
             print("Error: --refine and --focus are mutually exclusive.", file=sys.stderr)
             return 1
-        if not Path(raw_path).expanduser().resolve().is_dir():
+        if not raw_path or not Path(raw_path).expanduser().resolve().is_dir():
             print(
                 "Error: --refine requires an existing project directory, not a URL or idea.",
                 file=sys.stderr,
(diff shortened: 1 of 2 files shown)


shivchander:
@ceo-review

PR #1278, opened 2026-08-16
fix: tool engine auto-skips agent nodes when multiple nodes share the same role

### LINKED ISSUE #1277: Tool engine auto-skips agent nodes when multiple nodes share the same role
## Problem

`_detect_artifact()` in `factory/workflow/tool.py` auto-submits agent nodes by checking for `reviews/{role}-latest.md`. When multiple nodes in the same workflow share the same `AgentRole`, the first node's review file causes all subsequent same-role nodes to be auto-submitted without executing.

For example, if a workflow has two `BUILDER` nodes — one that produces artifact A and a later one that produces artifact B — the tool engine sees `reviews/builder-latest.md` from the first node and immediately marks the second node as completed. The second builder agent is never spawned, and artifact B is never produced.

## Root cause

In `_detect_artifact()`, the check order for `AgentNode` is:

1. Tagged review file: `reviews/{role}-{tag}-latest.md`
2. Generic review file: `reviews/{role}-latest.md`
3. Declared writes: files listed in `node.writes`

Check 2 matches before check 3 is reached. The generic review file has no way to distinguish which node wrote it.

## Impact

Any workflow with multiple nodes sharing the same agent role will have all but the first silently skipped by the tool engine. The `tool_next()` auto-submit loop advances past the node without the CEO ever being told to execute it.

shivchander:
Closes #1277

## Changes

- Reordered `_detect_artifact()` in `factory/workflow/tool.py` to check declared `node.writes` files **before** the generic `reviews/{role}-latest.md` fallback
- Added `return None` after the writes check to prevent fallthrough to the generic review file when writes are declared but absent — this is the key fix that prevents same-role collision
- Generic review file is now only checked for nodes with no declared writes and no tag match (backward compatible)
- Added 5 regression tests covering: same-role collision, writes priority over generic, backward compat for no-writes nodes, no-fallthrough when writes are absent, and the graph_explorer real-world scenario

### DIFF
--- a/factory/workflow/tool.py
+++ b/factory/workflow/tool.py
@@ -799,6 +799,7 @@ def _detect_artifact(
 
     if isinstance(node, AgentNode):
         role = node.role.value
+        # 1. Tagged review file
         tag = nid.replace(f"{role}_", "").replace(role, "")
         if tag and tag != nid:
             tagged_file = reviews_dir / f"{role}-{tag}-latest.md"
@@ -806,11 +807,7 @@ def _detect_artifact(
                 content = tagged_file.read_text().strip()
                 if content:
                     return content
-        review_file = reviews_dir / f"{role}-latest.md"
-        if review_file.exists() and _fresh(review_file):
-            content = review_file.read_text().strip()
-            if content:
-                return content
+        # 2. Declared writes — checked before generic to avoid same-role collision
         if node.writes:
             for wp in node.writes:
                 f = project_path / wp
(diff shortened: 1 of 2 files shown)


shivchander:
@ceo-review

shivchander:
## Factory CEO — KEEP Approval

**Experiment:** #6 | **Issue:** #1277 | **Verdict:** KEEP

### Scores
| Metric | Value |
|--------|-------|
| Before | 0.5774 |
| After | 0.633 |
| Delta | +0.0556 |

### QA Pipeline Results
- **Health Check:** PASS — composite +0.0556, lint clean, types clean, 84/84 targeted tests pass
- **Code Review:** CLEAN — 0 issues across 7 categories, spec fidelity 4/4
- **Adversarial Test:** PASS — 5 spec tests + 7 adversarial edge cases verified, 312/312 regression tests pass

### Change Summary
Reordered `_detect_artifact()` check logic in `factory/workflow/tool.py` to check `node.writes` before the generic `reviews/{role}-latest.md` fallback. Added `return None` barrier after writes check to prevent same-role node collisions. 5 regression tests added.

Aligns with how every major DAG engine (Airflow, Argo, Prefect, Dagster) handles node identity — artifacts keyed by instance, not by type.

PR #1283, opened 2026-08-16
fix: eliminate TOCTOU race in tmux-persist session creation (#1279)

### LINKED ISSUE #1279: Race condition in tmux-persist when spawning parallel agents
## Problem

When multiple agents are spawned in parallel via shell backgrounding (`&` + `wait`), the `run_in_tmux()` function in `factory/runners/_tmux_persist.py` has a race condition in session creation.

## Root cause

At lines 199-212, the code checks `_session_exists(session)` and branches:
- If session exists → `tmux new-window` (add a window to the existing session)
- If session doesn't exist → `tmux new-session` (create a new session)

When 3 agents are backgrounded simultaneously, all 3 check `_session_exists()` before any of them have completed `new-session`. All 3 see `has_session=False`, so all 3 try `tmux new-session` with the same session name. Only the first succeeds — the other two fail with "duplicate session" and return error code 1. The agents are never spawned.

## Impact

Parallel agent execution fails silently. Only the first agent in a parallel batch gets spawned. The others fail with misleading "duplicate session" errors. This was observed during design mode QA where 3 parallel agents (health_checker, code_reviewer, adversarial_tester) were spawned — only health_checker got the session.

Note: tagged parallel researchers work sometimes because they have a small startup delay between them, but the race condition is still present.

## Environment

- `factory/runners/_tmux_persist.py:158-212`
- Observed in design mode with parallel QA agents

shivchander:
Closes #1279

## Changes

- **Replace TOCTOU check-then-act with EAFP** in `factory/runners/_tmux_persist.py` lines 199-212: instead of calling `_session_exists()` then branching between `new-session` and `new-window`, try `tmux new-session` first and fall back to `tmux new-window` on non-zero returncode. The tmux server serializes concurrent `new-session` calls, so only one wins — the rest now gracefully fall back instead of failing.
- **Update `test_creates_new_session_when_none_exists`** — removed `_session_exists` mock; verifies `new-session` is attempted first.
- **Update `test_creates_window_when_session_exists`** — now tests the fallback path: `new-session` returns non-zero, `new-window` succeeds.
- **Update `test_returns_error_on_tmux_window_failure`** — both `new-session` and `new-window` must fail for the error path.
- **Add `test_race_condition_fallback_to_new_window`** — simulates the parallel agent race (duplicate session error on `new-session`), verifies fallback to `new-window` succeeds.
- **Remove dead `_session_exists` mocks** from all remaining tests (function is preserved for other callers but no longer called in `run_in_tmux`).

### DIFF
--- a/factory/runners/_tmux_persist.py
+++ b/factory/runners/_tmux_persist.py
@@ -196,17 +196,15 @@ async def run_in_tmux(
     )
     wrapper_script.chmod(0o755)
 
-    has_session = _session_exists(session)
-    if has_session:
+    result = subprocess.run(
+        ["tmux", "new-session", "-d", "-s", session, "-n", window,
+         "-x", "200", "-y", "50", str(wrapper_script)],
+        cwd=cwd,
+        capture_output=True,
+    )
+    if result.returncode != 0:
         result = subprocess.run(
             ["tmux", "new-window", "-t", session, "-n", window, str(wrapper_script)],
-            cwd=cwd,
-            capture_output=True,
-        )
-    else:
-        result = subprocess.run(
-            ["tmux", "new-session", "-d", "-s", session, "-n", window,
-             "-x", "200", "-y", "50", str(wrapper_script)],
             cwd=cwd,
             capture_output=True,
         )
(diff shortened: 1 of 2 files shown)


shivchander:
@ceo-review

PR #1293, opened 2026-08-17
fix: thread --focus through study subgraph and skill export (#1291)

### LINKED ISSUE #1291: Study subgraph does not pass --focus to study or graph_explorer
## Summary

The study subgraph (`_study_subgraph()` in `factory/workflow/definitions.py`) runs a general codebase analysis even when the user provides a `--focus` argument. Neither the `study` node nor the `graph_explorer` node receives the user's focus context.

## Current behavior

1. **`study` node** runs `factory study {project_path}` — no `--focus` flag, despite the CLI supporting `factory study <path> --focus <item>`
2. **`graph_explorer`** prompt says "Read observations.md for focus context" and use `"<focus from observations>"` — but this is instructional text, not a template variable. The agent derives focus from the broad study output, which may not reflect the user's intent

When a user runs `factory ceo /path --mode design --focus "auth flow"`, the focus flows through the CEO but never reaches the study subgraph. The study does a broad scan and the graph_explorer guesses the focus from that.

## Expected behavior

1. The `study` node should pass `--focus` when available: `factory study {project_path} --focus "{focus}"`
2. The `graph_explorer` prompt should receive the focus as a template variable so it can target graph queries precisely (e.g., `factory graph query "{project_path}" "auth flow" --depth 2` instead of guessing from observations)

## Impact

- Wasted graph exploration time on irrelevant parts of the codebase
- The downstream Researcher reads `study-combined.md` which may not emphasize the user's area of interest
- Particularly impactful on large codebases where a focused study would be significantly faster and more relevant

## Related

- #1256 — Researcher graph_explorer falls back to grep (fixed the path issue, this is about focus context)

mihirathale98:
Closes #1291

## Changes

- Add `_graph_explorer_prompt(focus)` helper that returns a focused prompt with the user's topic baked into graph query commands, falling back to `_GRAPH_EXPLORER_PROMPT` when focus is None
- Add `focus` keyword arg to `_study_subgraph()` — passes it to the `Study` node and uses the focused prompt for `graph_explorer`
- Update `_study_to_instruction()` to emit a CEO hint about `--focus` when `node.focus` is None (interactive path where the CEO reads the SKILL.md)
- Existing callers (`design_workflow`, `study_standalone_workflow`) unchanged — `focus=None` default preserves backward compatibility
- 7 new tests covering both the subgraph wiring and the skill export output

### DIFF
--- a/factory/workflow/definitions.py
+++ b/factory/workflow/definitions.py
@@ -45,6 +45,7 @@ __all__ = [
     "DOC_FRESHNESS_GATE_PROMPT",
     "ResearcherConfig",
     "_GRAPH_EXPLORER_PROMPT",
+    "_graph_explorer_prompt",
     "_research_subgraph",
     "_study_subgraph",
     "build_workflow",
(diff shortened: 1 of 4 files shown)


mihirathale98:
@ceo-review

PR #1297, opened 2026-08-17
fix: update test helper to write declared artifacts, preventing infinite loop

osilkin98:
Closes the infinite loop in `tests/test_loop_context.py` caused by commit f33b35b0's `_detect_artifact` change.

## Changes

- Updated `_simulate_reloop_cycle()` to write review files to each node's declared `writes` paths (e.g., `.factory/reviews/health-check.md` for `health_checker`) instead of always writing only the generic `{role}-latest.md` file
- Falls back to the generic path for nodes without declared writes, preserving backward compatibility
- No production code changes — only the test helper was modified

## Root Cause

After f33b35b0, `_detect_artifact()` returns `None` when a node has declared writes but those files don't exist on disk (no fallthrough to generic review file). The test helper only created the generic file, so `tool_next()`'s while loop never advanced past nodes like `health_checker` or `code_reviewer`.

## Verification

All 24 tests in `test_loop_context.py` pass in <1s (previously hung indefinitely). All 84 tests in `test_workflow_tool.py` also pass — no regressions.


osilkin98:
Closing — PR included 46k lines of unrelated upstream changes due to worktree base mismatch. Will reopen with a clean branch.

PR #1298, opened 2026-08-17
fix: update test helper to write declared artifacts, preventing infinite loop

osilkin98:
Closes #1296

## Root Cause

Commit f33b35b0 reordered `_detect_artifact` to check `node.writes` before the generic `reviews/{role}-latest.md` path. The `_simulate_reloop_cycle()` test helper in `TestLoopContextE2EComparison` was still writing only generic `{role}-latest.md` files for every `AgentNode`, regardless of what the node declared. Nodes with explicit `writes` (e.g. `health_checker` → `.factory/reviews/health-check.md`) never had their declared artifacts created, so `tool_next`'s artifact detection loop spun forever waiting for files that would never appear.

## Fix

When a node declares `writes`, write those exact paths instead of the generic fallback. Nodes without `writes` still get the old `{role}-latest.md` behavior.

## Verification

All 24 tests in `tests/test_loop_context.py` pass in ~0.3s, including the previously-hanging `test_ab_comparison_web_app`, `test_ab_comparison_library`, and `test_ab_report_generation`.

### DIFF
--- a/factory/workflow/contributed/outer_loop/test_workflow.py
+++ b/factory/workflow/contributed/outer_loop/test_workflow.py
@@ -22,7 +22,7 @@ class TestOuterLoopWorkflow:
 
     def test_node_count(self) -> None:
         wf = workflow()
-        assert len(wf.nodes) == 5
+        assert len(wf.nodes) == 6
 
     def test_required_nodes_present(self) -> None:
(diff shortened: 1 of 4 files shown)


osilkin98:
@ceo-review

PR #1299, opened 2026-08-17
ci: parallelize test suite with pytest-xdist

lambdabaa:
## Summary

- Add `pytest-xdist>=3.5` to dev dependencies
- Enable `-n auto` in CI test step for parallel test execution

## Why this is safe

Audited the full test suite (5,487 tests) for parallelism blockers:

| Check | Result |
|-------|--------|
| Bare `os.chdir` | None — all uses go through monkeypatch |
| Session/module-scoped fixtures | None — everything is function-scoped |
| Shared file paths | None — every test uses `tmp_path` |
| Port/socket binds | None |
| Direct `os.environ` mutation | 6 sites, but process-isolated under xdist |
| Global mutable state | Conftest resets are process-isolated |

## Results

Local run with `-n auto`: **5,470 passed in 5:10** (vs 10+ min sequential).
5 pre-existing failures unrelated to parallelism. On CI's 4-vCPU runners, expect 3-4x speedup.

## Test plan

- [x] Full suite passes with `-n auto` locally (5,470 passed, 5 pre-existing failures, 12 skipped)
- [x] Smoke test on subset (174 passed in 3.11s)
- [x] Pre-existing failures confirmed to reproduce without xdist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install dependencies
         run: uv sync --all-groups
       - name: Run tests with coverage
-        run: uv run pytest -v --tb=short --cov=factory --cov-report=xml
+        run: uv run pytest -n auto -v --tb=short --cov=factory --cov-report=xml
       - name: Upload coverage to Codecov
         if: matrix.python-version == '3.12'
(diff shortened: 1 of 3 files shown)


osilkin98:
@ceo-review

PR #1311, opened 2026-08-18
fix: conflict detector fails on branch checkout

lambdabaa:
## Summary

The PR Conflict Detector workflow fails at the "Commit conflicts.jsonl to conflict-data branch" step because `conflicts.jsonl` exists as an untracked file when `git checkout` tries to switch to the `conflict-data` branch.

```
error: The following untracked working tree files would be overwritten by checkout:
    conflicts.jsonl
```

**Root cause:** Line 31 creates `conflicts.jsonl` via `git show`, then line 184 tries to checkout `conflict-data` which also contains that file.

**Fix:** `rm -f conflicts.jsonl` before the checkout. The file is already saved to `/tmp/conflicts.jsonl.updated` at that point.

## Test plan

- [x] Verified the failure in [run 32162496911](https://github.com/akashgit/remote-factory/actions/runs/32162496911/job/95794395057)
- [x] Confirmed the fix: file is saved to /tmp before removal, restored after checkout


lambdabaa:
Closing — rebased as a clean single-commit PR to resolve docs/index.md conflict.

PR #1390, opened 2026-08-29
feat: guided KNOB_MUTATE with knob-contrastive reflection

lambdabaa:
## Summary

Extends the outer loop to optimize OptKnob values alongside graph topology. Previously, `apply_random_mutation` could only mutate graph structure (nodes, edges). Now it can also tune knob values declared by Packages, guided by contrastive analysis from the reflector.

### KNOB_MUTATE operator (`factory/outer_loop/mutations.py`)
- New mutation operator at 25% default weight
- Picks a knob, changes its value within declared bounds
- When bounds exhausted + `expandable=True`: calls `default_knob_expander` (Opus via CLI) to invent new values
- **Guided by reflector**: 70% picks the winning knob+value from contrastive analysis, 30% explores randomly
- `_parse_knob_suggestion()` extracts `(knob, value)` from reflector's `KNOB_MUTATE:` suggestions

### Knob-contrastive reflection (`factory/outer_loop/reflector.py`)
- `reflect()` accepts optional `knob_values_by_id: dict[str, dict[str, object]]`
- `_extract_knob_patterns()`: per-knob gradient (avg score per value across all individuals)
- Top-K vs bottom-K knob contrast: identifies which values consistently separate winners from losers
- Produces actionable suggestions: `"KNOB_MUTATE: tactical_style=mate_focused (avg +150) outperforms broad (+30) by 120"`

### Infrastructure
- `_deep_copy_workflow` preserves `knob_values`/`knob_bounds`/`knob_expandable` through graph mutations
- `_try_mutation` exempts `KNOB_MUTATE` from mutable_nodes guard
- `reflection_report` threaded through `apply_random_mutation` → `_try_mutation` → `mutate_knob`

### Docs
- CLAUDE.md, SPEC.md, outer-loop.md: 7→8 operators
- workflow/README.md: Package ecosystem section
- reflect.md, workflow-tune.md: KNOB_MUTATE examples

## Validated by

Chess evolution demo: factory's reflector identified `board_representation=both outperforms ascii by 118 points` and `pipeline_mode=enumerate outperforms parallel by 13`. Guided KNOB_MUTATE used these gradients to preferentially select winning values.

## Test plan

- [x] 402 tests pass
- [x] Lint clean (ruff)
- [x] Type clean (mypy)
- [x] KNOB_MUTATE within bounds tested
- [x] Expander callback tested
- [x] Guided mutation from reflection tested
- [x] Knob preservation through graph mutations tested

[...]

PR #1407, opened 2026-08-30
fix: include prompt content in MAP-Elites feature extraction

### LINKED ISSUE #1388: compute_features ignores OptKnob values — MAP-Elites grid collapses for knob-level optimization
## Problem

`compute_features()` in `factory/outer_loop/similarity.py` extracts only topology dimensions (depth, fork_degree, agent_count, gate_count). When the optimization surface is primarily `OptKnob` values (prompt styles, thresholds, boolean flags) rather than graph structure changes, every candidate maps to the same grid cell and the MAP-Elites archive degenerates to a single-best selector.

This is a pre-existing limitation in the outer loop — `compute_features` was written before the Package ecosystem and only knows about raw workflow graphs. The chess evolution demo (#1384) made it visible as the first use case that optimizes knob values through the outer loop.

## Expected behavior

`compute_features` should inspect `OptKnob` declarations on the workflow's source `Package` and include categorical knob values as additional feature dimensions. Any Package with declared knobs should get diversity pressure across its optimization surface without the user writing a custom feature function.

## Workaround

The chess demo defines a manual `chess_features(cfg)` function that maps 4 key knobs to grid axes, then overrides `ind.features` after `Population.make_individual()`. This works but defeats the purpose of the declarative `OptKnob` abstraction.

## Root cause

Gap between the Package ecosystem (which treats knobs as first-class declarative metadata) and the outer loop (which only sees compiled graph structure). `Package.compile()` flattens knob values into node parameters, losing the knob identity that `compute_features` would need.

Ref: #1384 (Package ecosystem prototype — where this was discovered)

lambdabaa:
## Summary

Fixes #1388.

- `compute_features()` now hashes each `AgentNode.prompt_template` into a feature dimension (8 buckets per node)
- Without this, every `PROMPT_MUTATE` that doesn't also change a knob lands in the **same MAP-Elites cell** — only the highest-scoring prompt survives, and all other prompt diversity is discarded
- This is the `PROMPT_MUTATE` counterpart to the knob-hash fix we landed in #1390

Discovered while running the chess evolution demo: with 50% `PROMPT_MUTATE` weight, most prompt experiments were being thrown away because they all mapped to the same cell.

## Test plan

- [x] `test_simple_workflow` — features include prompt dimensions
- [x] `test_workflow_with_fork` — 3 agent prompt features appended
- [x] All 46 similarity/feature tests pass
- [x] `ruff check` clean
- [x] `mypy` clean

### DIFF
--- a/factory/outer_loop/similarity.py
+++ b/factory/outer_loop/similarity.py
@@ -73,10 +73,13 @@ def graph_edit_distance(w1: Workflow, w2: Workflow) -> int:
 def compute_features(workflow: Workflow) -> tuple[int, ...]:
     """Extract features from a workflow for MAP-Elites grid placement.
 
-    Base features: (depth, fork_degree, agent_count, gate_count).
-    If the workflow carries knob_values (from Package.compile()), each
-    categorical knob value is hashed to an integer bucket and appended,
-    giving the archive diversity pressure across the optimization surface.
+    Features: (depth, fork_degree, agent_count, gate_count,
+               *knob_hashes, *prompt_hashes).
+
+    Knob values (from Package.compile()) and agent prompt_templates are
+    each hashed to integer buckets so that KNOB_MUTATE and PROMPT_MUTATE
+    produce distinct cells — without this, prompt-only mutations collapse
+    into a single cell and only the best survives.
     """
     g = _build_nx_graph(workflow)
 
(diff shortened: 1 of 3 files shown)


lambdabaa:
@ceo-review

lambdabaa:
@ceo-review

PR #1409, opened 2026-08-30
fix: add observability to prompt rewriter and knob expander

lambdabaa:
## Summary

- **Timeout bump**: 60s → 120s for both `default_prompt_rewriter` and `default_knob_expander` CLI calls. 60s was causing silent failures — Opus needs time for quality prompt rewrites.
- **Structured logging**: Every failure path now emits a structlog event:
  - `prompt_rewritten` / `prompt_rewriter_empty` / `prompt_rewriter_timeout` / `prompt_rewriter_error`
  - `prompt_mutate_validation_failed` (when rewritten prompt fails Pydantic model_copy)
  - `knob_expanded_via_cli` / `knob_expander_empty` / `knob_expander_timeout` / `knob_expander_error`

Previously, all `except Exception: return None` paths swallowed errors silently, making it impossible to diagnose why `PROMPT_MUTATE` succeeded at a lower rate than its configured weight.

Discovered while running the chess evolution demo — prompt mutations were ~17% of actual mutations despite 50% weight. The logging confirmed the rewriter works (20/20 successes) but the missing observability would have left us guessing.

## Test plan

- [x] `ruff check` clean
- [x] `mypy` clean
- [x] Verified in chess demo: 20 consecutive `prompt_rewritten` events, 0 timeouts/errors

### DIFF
--- a/factory/outer_loop/mutations.py
+++ b/factory/outer_loop/mutations.py
@@ -552,11 +552,19 @@ def default_prompt_rewriter(
             ["claude", "-p", prompt, "--model", "opus",
              "--append-system-prompt", "Output only the prompt text.",
              "--max-turns", "1", "--output-format", "text"],
-            capture_output=True, text=True, timeout=60,
+            capture_output=True, text=True, timeout=120,
         )
         result = proc.stdout.strip()
+        if result:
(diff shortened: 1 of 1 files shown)


lambdabaa:
@ceo-review

PR #1411, opened 2026-08-30
fix: PROMPT_MUTATE persists in knob_values for compile() round-trips

### LINKED ISSUE #1410: Mutated Workflow IR should be directly evaluatable without config round-trip
## Problem

`apply_random_mutation()` returns a mutated `Workflow` IR with changes applied (e.g., rewritten `prompt_template` on an `AgentNode`). But consumers that rebuild from a config object (like `Package.compile()`) lose these mutations because `compile()` always starts fresh from the Package's knobs/defaults.

This means **PROMPT_MUTATE has no effect** in any outer loop that follows the pattern:
```python
child_wf, rec = apply_random_mutation(parent_wf, strategy, gen)
# ... extract knob overrides from child_wf.knob_values ...
child_pipeline = build_pipeline(cfg_with_overrides)  # rebuilds from config
result = evaluate(child_pipeline)  # prompt mutation is lost
```

The mutated prompt lives in `child_wf.nodes[target].prompt_template` but is never seen by the evaluator.

## Impact

Discovered in the chess evolution demo: after 30+ generations with 50% PROMPT_MUTATE weight and a working `default_prompt_rewriter` (38/38 Opus rewrites succeeded), every prompt mutation was actually evaluated with the **original** prompt. The rewritten prompts were silently discarded.

This affects any outer loop consumer that doesn't evaluate the `Workflow` IR directly — which is the natural pattern when using Package ecosystem's `compile()`.

## Proposed fix

The outer loop should carry prompt (and other node-level) mutations through the evaluation path. Options:

1. **`Workflow.apply_to(package: Package) -> Package`** — patches mutations from the IR back onto a Package, returning a new Package ready to compile+evaluate
2. **`WorkflowExecutor` accepts `Workflow` directly** — skip the Package round-trip entirely for evaluation
3. **Store node-level mutations in `knob_values`** — treat prompts as knobs so they survive the `compile()` round-trip (this is what `OptKnob(kind="prompt")` was designed for, but `mutate_prompt` bypasses it)

Option 3 is the most aligned with the existing design — `PROMPT_MUTATE` should write to `knob_values` when the target node has an associated `OptKnob(kind="prompt")`, so mutations survive `compile()`.

## References

- PR #1403 (free-form prompt rewriting)
- PR #1407 (MAP-Elites feature extraction)
- Package ecosystem: `factory/workflow/package.py`

lambdabaa:
## Summary

Fixes #1410.

`mutate_prompt()` now stores the rewritten prompt in `knob_values` under a synthetic `_prompt_<node_id>` key. `Package.compile()` reads these back and applies them to node `prompt_templates`.

**Without this fix, PROMPT_MUTATE has no effect** in any outer loop that rebuilds from config:

```python
child_wf, rec = apply_random_mutation(parent_wf, ...)  # prompt rewritten on IR
child_pipeline = build_pipeline(cfg)  # rebuilds from config, prompt lost
result = evaluate(child_pipeline)      # evaluates with ORIGINAL prompt
```

Discovered in the chess demo: after 30+ generations with 50% PROMPT_MUTATE weight and 38/38 successful Opus rewrites, every prompt mutation was evaluated with the original prompt. The rewritten prompts were silently discarded.

## Changes

- `mutate_prompt()` writes `_prompt_<node_id> = new_prompt` to `wf.knob_values`
- `Package.compile()` applies `_prompt_*` knobs back to nodes after compilation
- Prompt mutations now survive the `config → Package → compile()` round-trip

## Test plan

- [x] All 71 mutation + package ecosystem tests pass
- [x] End-to-end round-trip test: mutate → store in knob_values → compile() → prompt applied
- [x] `ruff check` clean
- [x] `mypy` clean

### DIFF
--- a/factory/outer_loop/mutations.py
+++ b/factory/outer_loop/mutations.py
@@ -599,8 +599,14 @@ def mutate_prompt(
     try:
         updated = node.model_copy(update={"prompt_template": new_prompt})
         wf.nodes[node_id] = updated  # type: ignore[assignment]
-    except Exception:
+    except Exception as exc:
+        log.warning("prompt_mutate_validation_failed", node=node_id, error=str(exc))
         return None
+
+    # Persist in knob_values so the mutation survives Package.compile() round-trips
+    prompt_knob = f"_prompt_{node_id}"
+    wf.knob_values[prompt_knob] = new_prompt
+    wf.knob_expandable[prompt_knob] = f"Prompt for {node_id}"
 
     record = MutationRecord(
         operator=MutationType.PROMPT_MUTATE,
--- a/factory/workflow/package.py
+++ b/factory/workflow/package.py
@@ -137,12 +137,23 @@ class Package(BaseModel):
         survive compile() round-trips.
         """
         wf = self.graph.model_copy(deep=True)
+        # Preserve _prompt_* entries from previous mutations before overwriting
+        saved_prompts = {
+            k: v for k, v in wf.knob_values.items()
+            if k.startswith("_prompt_")
+        }
+        saved_expandable = {
+            k: v for k, v in wf.knob_expandable.items()
+            if k.startswith("_prompt_")
+        }
         if self.knobs:
             wf.knob_values = {k.name: k.default for k in self.knobs}
             wf.knob_bounds = {k.name: list(k.bounds) for k in self.knobs if k.bounds}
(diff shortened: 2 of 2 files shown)


lambdabaa:
@ceo-review

lambdabaa:
@ceo-review

PR #1394, opened 2026-08-29
feat: rank-weighted tournament selection for MAPElitesArchive

lambdabaa:
## Summary

`sample_parent()` now supports rank-weighted sampling: individuals are drawn with probability proportional to their rank (best=N, worst=1) instead of uniformly. This biases toward stronger parents while preserving diversity.

## Problem

With uniform tournament selection, a -1564 score individual has the same probability of being drawn as a +123 one. In the chess evolution demo, 82% of archive cells scored negative — uniform k=3 tournaments frequently picked bad parents, wasting mutations.

## Changes

- `MAPElitesArchive.sample_parent(rank_weighted=True)` — uses `random.choices` with rank-proportional weights
- `SwarmConfig.rank_weighted_selection: bool = False` — backward compatible config option
- `SwarmEngine` threads the config through to `sample_parent`

## Test plan

- [x] 23 population tests pass (2 new)
- [x] Rank-weighted bias verified: best individual selected >33% with k=1 (vs ~33% uniform)
- [x] Uniform baseline unchanged when rank_weighted=False
- [x] Lint + type clean

Fixes #1393

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1412, opened 2026-08-30
feat: auto-switch to rank-weighted selection based on archive state

lambdabaa:
## Summary

Builds on #1394 (rank-weighted tournament selection). `sample_parent()` now auto-activates rank-weighted selection when the archive reaches `auto_rank_cell_threshold` (default 8) occupied cells, indicating enough diversity that biasing toward stronger parents is beneficial.

This mirrors the `on_plateau()` pattern — adapting selection pressure based on search state rather than requiring static configuration. Auto-switch is on by default; callers can opt out with `auto_rank_weighted=False`.

Cell count is the sole trigger — score variance was considered but dropped because a fixed variance threshold is application-dependent (chess scores range -500 to +500, factory's default scores are 0-1 floats). Cell count is application-agnostic: enough diverse individuals in the archive → rank-weighted helps regardless of score scale.

## Test plan

- [x] `test_auto_activates_by_cell_count` — triggers at 10 cells with threshold 8
- [x] `test_auto_stays_uniform_below_thresholds` — 2 cells → stays uniform
- [x] `test_auto_disabled` — `auto_rank_weighted=False` keeps uniform
- [x] All 26 population tests pass

### DIFF
--- a/factory/outer_loop/population.py
+++ b/factory/outer_loop/population.py
@@ -129,6 +129,8 @@ class MAPElitesArchive:
         self,
         tournament_size: int = 3,
         rank_weighted: bool = False,
+        auto_rank_weighted: bool = True,
+        auto_rank_cell_threshold: int = 8,
     ) -> Individual | None:
         """Tournament selection: pick tournament_size individuals, return the best.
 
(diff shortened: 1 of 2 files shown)


lambdabaa:
@ceo-review

PR #1414, opened 2026-08-31
fix: eliminate no-op knob mutations

lambdabaa:
## Summary

Two fixes to `mutate_knob()` that waste evaluation slots:

1. **Guided no-op detection**: when the reflector suggests a value that matches the current value (`theory -> theory`, `False -> False`), skip it and fall through to random selection instead of evaluating an identical config.

2. **Exclude `_prompt_*` synthetic knobs**: these are created by `PROMPT_MUTATE` (#1411) to persist prompts through `compile()` round-trips. `KNOB_MUTATE` should not randomly select them since they contain full prompt text, not tunable parameter values.

Discovered in the chess demo: ~15% of knob mutations were no-ops (`theory -> theory`, `2.0 -> 2.0`, `False -> False`), wasting evaluation budget on configs identical to the parent.

## Test plan

### DIFF
--- a/factory/outer_loop/mutations.py
+++ b/factory/outer_loop/mutations.py
@@ -724,8 +724,16 @@ def mutate_knob(
                 new_val = float(new_val)
             except ValueError:
                 pass
-    else:
-        knob_name = random.choice(knob_names)
+        # Skip no-op: guided value same as current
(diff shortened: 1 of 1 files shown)


lambdabaa:
@ceo-review

PR #1422, opened 2026-08-31
perf: add --bare flag to agent subprocess calls

### LINKED ISSUE #1420: perf: use --bare flag for agent subprocess calls
Factory spawns claude subprocesses for each agent invocation without \`--bare\`, meaning every call loads hooks, LSP, and plugins. For the outer loop which spawns hundreds of agent calls, this adds 1-2s of overhead per call.

\`claude -p ... --bare\` skips non-essential startup and should be the default for programmatic/headless agent invocations.

Discovered in chess-evolve demo: 225 subprocess spawns per variant, ~2s overhead each = ~7.5 minutes of pure startup overhead per variant.

lambdabaa:
## Summary

Fixes #1420.

Adds `--bare` to both `_build_headless_cmd` methods in `factory/runners/claude.py`. This skips hooks, LSP, and plugin loading on every agent subprocess invocation, saving ~1-2s of startup overhead per call.

Discovered in the chess-evolve demo: 225 subprocess spawns per variant × ~2s overhead = ~7.5 minutes of pure startup overhead per variant.

## Test plan

- [x] `--bare` is compatible with `--verbose` and other flags

### DIFF
--- a/factory/runners/claude.py
+++ b/factory/runners/claude.py
@@ -116,6 +116,7 @@ class ClaudeRunner:
             "--output-format",
             "stream-json",
             "--verbose",
+            "--bare",
             "--disallowedTools",
             "Agent",
         ]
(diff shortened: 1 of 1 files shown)


lambdabaa:
@ceo-review

PR #1423, opened 2026-08-31
feat: reflector includes prompt mutations in contrastive analysis

### LINKED ISSUE #1418: Reflector cannot see PROMPT_MUTATE content for contrastive analysis
The \`OuterLoopReflector\` does knob-contrastive analysis via \`knob_values_by_id\`, but \`PROMPT_MUTATE\` operates on the workflow IR's \`prompt_template\` field, not on knob values. The reflector has no visibility into what prompt text changed between winners and losers.

This means the reflector can suggest \"try a different prompt on tactician\" but cannot identify *which* prompt changes helped or hurt. It sees the score difference but not the causal prompt difference.

Proposed fix: \`mutate_prompt\` already stores prompts in \`_prompt_*\` knob values (#1411). The reflector should automatically include \`_prompt_*\` entries in its contrastive analysis, comparing prompt content (or a hash/summary) between top-K and bottom-K individuals.

lambdabaa:
## Summary

Fixes #1418.

`_extract_knob_patterns` now recognizes `_prompt_*` and `prompt_*` keys in `knob_values_by_id`:
- Truncates prompt values to 100 chars for comparison (full text would make every value unique)
- Reports `PROMPT_MUTATE` suggestions instead of `KNOB_MUTATE` for prompt-type keys
- Displays truncated values (60 chars) in suggestions

Without this, the reflector can see that a prompt mutation happened but can't compare prompt content between winners and losers.

## Test plan

- [x] All 7 reflector tests pass

### DIFF
--- a/factory/outer_loop/reflector.py
+++ b/factory/outer_loop/reflector.py
@@ -240,12 +240,16 @@ class OuterLoopReflector:
             all_knob_names.update(kv.keys())
 
         for knob in sorted(all_knob_names):
+            is_prompt = knob.startswith("_prompt_") or knob.startswith("prompt_")
             val_scores: dict[str, list[float]] = {}
             for id_, score, _ in all_sorted:
                 kv = knob_values_by_id.get(id_, {})
                 val = str(kv.get(knob, ""))
-                if val:
-                    val_scores.setdefault(val, []).append(score)
+                if not val:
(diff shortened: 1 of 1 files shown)


lambdabaa:
@ceo-review

PR #1424, opened 2026-08-31
fix: don't halt workflow on agent non-zero exit code

### LINKED ISSUE #1415: Executor halts entire workflow on invoke_agent return code 1
Any custom `invoke_agent` implementation that returns exit code 1 (e.g., empty LLM response) causes the executor to set `result.halted = True` and abort all remaining nodes, including nodes in other fork branches.

This means a single transient API failure kills the entire pipeline run. The executor should distinguish between retryable agent failures (bad output, timeout) and fatal process crashes.

Discovered in chess-evolve demo: every game was silently halted because the custom invoke_agent returned code 1 on empty responses.

lambdabaa:
## Summary

Fixes #1415.

`_run_agent` was raising `RuntimeError` when `invoke_agent` returned a non-zero exit code, which caused the executor to set `result.halted = True` and abort all remaining nodes (including nodes in other fork branches).

Now logs a warning and returns the stdout instead. Non-zero exit codes are treated as degraded output, not fatal failures. The workflow continues with whatever the agent produced.

Discovered in chess-evolve: any custom `invoke_agent` that returned code 1 (e.g., on empty LLM response) killed the entire pipeline run.

## Test plan

- [x] All 39 executor tests pass

### DIFF
--- a/factory/workflow/executor.py
+++ b/factory/workflow/executor.py
@@ -843,7 +843,12 @@ class WorkflowExecutor:
         )
 
         if code != 0:
-            raise RuntimeError(f"agent {node.role.value} exited with code {code}")
+            log.warning(
+                "agent_nonzero_exit",
+                role=node.role.value,
+                code=code,
+                output_len=len(stdout),
+            )
 
         return stdout
 


lambdabaa:
@ceo-review

PR #1426, opened 2026-08-31
feat: add async_prompt_rewriter for non-blocking prompt rewriting

### LINKED ISSUE #1417: default_prompt_rewriter blocks the async event loop
\`default_prompt_rewriter\` uses \`subprocess.run\` (synchronous), which blocks the event loop when called from async code via \`apply_random_mutation\`. Any async consumer that calls mutation operators gets blocked for 30-120s per prompt rewrite.

The rewriter should use \`asyncio.create_subprocess_exec\` or factory should provide an async variant. Callers currently have to wrap in \`asyncio.to_thread\`, which spawns uncontrolled threads and bypasses any async semaphores.

lambdabaa:
## Summary

Fixes #1417.

Adds `async_prompt_rewriter` alongside the existing sync `default_prompt_rewriter`. Uses `asyncio.create_subprocess_exec` instead of `subprocess.run`, so async consumers don't block the event loop for 30-120s per prompt rewrite.

The sync version remains as the default for backward compatibility. Async consumers can pass `async_prompt_rewriter` as the rewriter parameter to `mutate_prompt`.

## Test plan

### DIFF
--- a/factory/outer_loop/mutations.py
+++ b/factory/outer_loop/mutations.py
@@ -568,6 +568,51 @@ def default_prompt_rewriter(
         return None
 
 
+async def async_prompt_rewriter(
+    node_id: str,
+    current_prompt: str,
+    hint: str | None,
+) -> str | None:
+    """Async prompt rewriter: uses claude CLI without blocking the event loop."""
+    import asyncio as _asyncio
+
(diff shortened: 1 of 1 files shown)


lambdabaa:
@ceo-review

PR #1440, opened 2026-09-02
 gate --bare on API-key-style auth to fix subscription login

### LINKED ISSUE #1437: --bare flag (#1420/#1422) breaks Claude Code subscription (OAuth) auth  contradicts "no API key needed" docs
## Summary

PR #1422 (fixing #1420) added `--bare` to every `claude` subprocess spawned by `ClaudeRunner` in `factory/runners/claude.py` — both `build_command()` (headless) and `build_interactive_command()` (interactive).

This breaks `factory ceo` for anyone authenticated via a Pro/Max subscription login. Per `claude --bare`'s own `--help` text, it skips **keychain reads** — which is where subscription OAuth credentials live. The identical call without `--bare` works fine.

This directly contradicts `docs/setup.md`:

> "No API keys for re:factory's Claude Code subprocesses — however you've authenticated Claude Code is how re:factory accesses Claude. No separate Anthropic API key is needed."

That claim only holds for `ANTHROPIC_API_KEY` / Vertex / Bedrock auth. It does **not** hold for subscription/OAuth auth, because of `--bare`.

## Repro

1. Authenticate Claude Code via subscription login (`claude` → `/login`), not `ANTHROPIC_API_KEY`.
2. Run:
   ```
   factory install && factory ceo <project> --mode design --headless --auto-approve --no-github
   ```
3. Every CEO subprocess spawn fails instantly with exit code 1:
   ```
   [warning] claude_nonzero_exit  code=1 stderr=
   ceo agent exited with code 1
   ...
   Error: Aborting after 2 consecutive agent spawn failures. Last failed agent: ceo.
   ```
   The error message doesn't point to the actual cause.
4. Manually isolate:
   - `claude -p "say hi"` → succeeds
   - `claude --bare -p "say hi"` → `Not logged in · Please run /login`
5. Confirm the auth-method dependency: setting a dummy `ANTHROPIC_API_KEY` and rerunning `claude --bare -p "say hi"` no longer fails instantly — it proceeds to attempt a real API call. This confirms `--bare` only works with API-key-style auth (it skips keychain/OAuth entirely).

## Suggested fix

Don't unconditionally add `--bare`. It's a legitimate perf win for API-key/Vertex/Bedrock users (the case #1420 was measuring: hundreds of subprocess spawns in an outer-loop benchmark) — but it shouldn't be unconditional.

Gate it on auth method: only pass `--bare` when `ANTHROPIC_API_KEY` / `CLAUDE_CODE_*` env vars are present; skip it otherwise so keychain-based/subscription logins keep working.

At minimum, update `docs/setup.md` to state that `--bare`-based headless/design-mode runs require API-key auth — right now subscription users are silently broken with no indication why.

## Environment

- `remote-factory` installed via `uv tool install git+https://github.com/akashgit/remote-factory.git`
- Auth: Claude Code subscription login (Pro/Max), not `ANTHROPIC_API_KEY`

ramansrivastava:
## Summary

- `--bare` explicitly skips OAuth keychain reads (per `claude --help`), so #1420/#1422
  adding it unconditionally to every Claude Code subprocess broke subscription (Pro/Max)
  logins — every agent spawn failed instantly with `Not logged in · Please run /login`.
- Adds `_supports_bare_mode()` in `factory/runners/claude.py`, which only passes `--bare`
  when `ANTHROPIC_API_KEY` / `CLAUDE_CODE_USE_VERTEX` / `CLAUDE_CODE_USE_BEDROCK` is truthy.
  Subscription users keep working (falling back to pre-#1422 behavior); API-key/Vertex/
  Bedrock users keep the startup perf win.
- Updated `docs/setup.md` to document this behavior next to the auth env var table, and
  added a previously-undocumented `CLAUDE_CODE_USE_BEDROCK` var to that table.
- Matches the fix suggested in the issue.

Fixes #1437

## Test plan

- `ruff check .` / `mypy factory/` clean
- New `TestBareMode` in `test_runners.py` — unit tests on `_supports_bare_mode` (API key /
  Vertex / Bedrock / unset / falsy values) plus both command builders (`build_command`,
  `build_interactive_command`) asserting `--bare` is included/omitted correctly
- Full `pytest` suite: 5275 passed, 18 failed — all 18 pre-existing/environment-specific
  (verified against `main`: missing `oc` CLI / k8s cluster, low disk space, one unrelated
  `create_v2` import lint failure), none touch the changed files
- Live verification against a real subscription login: `factory ceo ... --mode design
  --just-plan` completed normally with `"apiKeySource":"none"` in the init line,
  no  `Not logged in` error — reproduced the failure pre-fix, confirmed fixed post-fix

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -106,6 +106,12 @@ re:factory inherits Claude Code's authentication. Configure whichever method you
 | `CLAUDE_CODE_USE_VERTEX` | Set to `1` for Google Cloud Vertex AI |
 | `ANTHROPIC_VERTEX_PROJECT_ID` | Vertex AI project ID |
 | `CLOUD_ML_REGION` | Vertex AI region (e.g., `us-east5`) |
+| `CLAUDE_CODE_USE_BEDROCK` | Set to `1` for AWS Bedrock |
+
+When one of the variables above is set, re:factory's Claude Code subprocesses launch with
+`--bare` for faster startup (skips hooks, LSP, and plugin loading). `--bare` never reads the
+OAuth keychain, so it's skipped automatically for subscription (Pro/Max) logins — those keep
+working exactly as `claude` does on its own, just without the startup optimization.
 
 ### re:factory Configuration
 
--- a/factory/runners/claude.py
+++ b/factory/runners/claude.py
@@ -20,6 +20,25 @@ if TYPE_CHECKING:
 
 log = structlog.get_logger()
 
+# Boolean-flag env vars that switch Claude Code onto a 3P backend (their own credentials,
+# never the OAuth keychain), as opposed to ANTHROPIC_API_KEY which carries the key itself.
+_BARE_COMPATIBLE_BACKEND_FLAGS = ("CLAUDE_CODE_USE_VERTEX", "CLAUDE_CODE_USE_BEDROCK")
+
+
+def _supports_bare_mode(env: dict[str, str]) -> bool:
+    """True when the environment implies API-key-style auth, not OAuth/keychain login.
+
+    `--bare` skips keychain reads entirely (see `claude --help`), so it only works when
+    auth is ANTHROPIC_API_KEY or a 3P backend (Vertex/Bedrock). Passing it unconditionally
+    breaks subscription (Pro/Max) logins, which live in the keychain — see #1437.
+    """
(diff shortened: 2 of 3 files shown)


lambdabaa:
@ceo-review

PR #1467, opened 2026-09-07
fix: close outer loop reflection→mutation wiring gaps

colehurwitz:
Closes the 7 confirmed wiring gaps that made the CLI-driven outer loop's guided mutation pipeline structurally dead.

## Changes

- **`_cmd_evolve` loads persisted ReflectionReport** — deserializes `reflections/gen{N}.json` and passes it to `apply_random_mutation()`, activating guided mutations on the CLI path
- **`_cmd_reflect` passes `knob_values_by_id`** — builds knob values from registry workflows and passes to `reflector.reflect()`, activating `_extract_knob_patterns` (68 lines) on CLI path
- **`engine.py` passes `knob_values_by_id`** — builds knob values from population individuals for the in-process `SwarmEngine` path
- **Populate `prompt_improvements` in reflector** — `_extract_knob_patterns` now appends targeted prompt-improvement strings for prompt knobs, so `_extract_prompt_hint` returns actionable guidance instead of falling through to generic `success_patterns`
- **Domain-neutral `_PROMPT_VARIANTS`** — replaced coding-specific strings ("failing tests", "codebase", "tests first") with domain-neutral reasoning strategies
- **Fix dead `seed_workflow` config** — cleared `name = "improve"` to `name = ""` in both `featurebench.toml` and `swebench.toml` (not a registered workflow name, silently fell back)
- **Delete orphaned `prompts/reflect.md`** — unreferenced stub file

## Tests

- `test_evolve_loads_persisted_reflection_json` — seeds a reflection JSON fixture, spies on `apply_random_mutation`, asserts `reflection_report` is non-None
- `test_evolve_works_without_reflection_file` — verifies graceful fallback when no reflection exists
- `test_reflect_with_knob_values_produces_suggestions` — asserts `KNOB_MUTATE`/`PROMPT_MUTATE` suggestions and non-empty `prompt_improvements`
- `test_reflect_without_knob_values_no_knob_suggestions` — confirms no knob suggestions without `knob_values_by_id`
- `test_guided_operator_selection_with_reflection` — mocks `random.random` < 0.7 to verify guided path fires
- `test_no_coding_specific_keywords` — asserts `_PROMPT_VARIANTS` are domain-neutral

PR #1466, opened 2026-09-06
fix: persist eval_details and instance_results through disk round-trip

colehurwitz:
Closes #1465

## Changes

- **`factory/inner_loop.py`** — `_write_cycle_summary()` now accepts `instance_results`, `kept`, `reverted`; builds verify block via `verify_adapter.eval_result_from_verify_results()`; `_step_with_task()` calls it with instance_results so cycle_summary.json exists on disk for task-driven evaluations
- **`factory/outer_loop/evaluator.py`** — `CycleRecordCache.save_cache()/load_cache()` now persist and restore `eval_details` and `instance_results` in JSONL entries
- **`factory/cli/outer_loop.py`** — `_load_cycle_summary()` reconstructs `CycleRecord.eval_details` from `verify`, `test_details`, `rejected`, `error` keys and populates `instance_results`
- **`tests/test_outer_loop/test_verify_disk_roundtrip.py`** — Integration test: write → read → reflect pipeline producing non-empty failure/success patterns from verify data; CycleRecordCache round-trip tests

## Test plan

- [x] 5 new integration tests pass (disk round-trip + cache round-trip)
- [x] 92 existing outer loop tests pass (zero regressions)
- [x] `ruff check` clean
- [x] `mypy` clean

PR #1472, opened 2026-09-07
feat: outer loop optimization — LLM reflection, typed mutations, Task/Workflow separation

### LINKED ISSUE #1476: bug: outer-loop CLI evaluate does not propagate scores to population.json, and evolve has no parent selection
## File to modify

`factory/cli/outer_loop.py`

## Bug 1: `_cmd_evaluate` never writes scores back to population.json

**Location:** `_cmd_evaluate()`

**Problem:** The function evaluates each candidate and writes scores to `results/genN.json`, but never updates `population/population.json`. All individuals in the population retain `score=0.0` after evaluation.

**Impact:** The evolve step cannot distinguish good parents from bad ones. Tournament selection (in `SwarmEngine.run_generation`) uses population scores, and the CLI evolve path also needs scores to make informed parent selection.

**Fix:** After writing `results/genN.json` and before updating the checkpoint, load the `Population` from `population/population.json`, update each individual's score from the results dict, and save the population back. Mode names follow `evolve-gen{N}-{individual_id[:8]}`, so the mode-name suffix matches against individual IDs.

## Bug 2: `_cmd_evolve` has no parent selection — mutates all modes equally

**Location:** `_cmd_evolve()`

**Problem:** The function iterates `modes[:config.population_size]` and applies a random mutation to each mode. Every individual reproduces exactly once regardless of fitness. There is zero selection pressure — the worst candidate gets the same number of offspring as the best.

`SwarmEngine.run_generation()` does proper tournament selection via `self._archive.sample_parent()`, but the CLI path completely bypasses this.

**Fix:** Load scores from `results/genN.json`, filter modes to the current generation (prefix `evolve-gen{generation}-`), and implement k=2 tournament selection for each offspring slot. Print the parent's mode name and score for observability.

## Tests

Add tests in `tests/test_outer_loop/test_cli.py` covering:
1. After `_cmd_evaluate`, population.json has non-zero scores matching results
2. `_cmd_evolve` with score-differentiated parents produces offspring biased toward higher-scoring parents (statistical test)

## Constraints

- Do NOT change function signatures in a breaking way
- Do NOT modify `engine.py` or `population.py` — the fix is entirely in `cli/outer_loop.py`

colehurwitz:
## Architecture: Task → InnerLoop → OuterLoop

This PR establishes clean separation between **what to evaluate** (Task) and **what to evolve** (Workflow). Two CLI flags, one per concern:

```bash
factory outer-loop calibrate /path/to/project \
  --task-module chess_evolve.task:ChessEvolveTask \
  --seed-workflow chess_evolve.pipeline:build_pipeline
```

- `--task-module` → **domain** concern. A `Task` subclass with four hooks:
  - `instances()` — what to work on (chess positions, SWE-bench issues, molecules)
  - `setup(instance, workspace)` — prepare the environment
  - `prompt(instance)` — what should the agent do
  - `verify(instance, workspace)` — did it work? Returns `VerifyResult(passed, score, details)`

- `--seed-workflow` → **topology** concern. A callable returning a `Package` or `Workflow` with `OptKnob`s:
  - Defines the agent pipeline topology (which agents, in what order)
  - Declares tunable parameters via `OptKnob` (model, timeout, retry count, prompt style)
  - The outer loop mutates this topology + knobs across generations

### How execution flows

```
OuterLoop (MAP-Elites evolutionary search)
  │
  ├─ calibrate: load seed workflow from --seed-workflow, create initial population
  ├─ evaluate: for each candidate workflow in the population:
  │     │
  │     └─ InnerLoop._step_with_task():
  │           for each instance from task.instances():
  │             1. task.setup(instance, workspace)
  │             2. WorkflowExecutor(candidate_workflow, workspace,
  │                                 initial_context=task.prompt(instance))
  │             3. executor.execute()  — walks the workflow graph deterministically
  │             4. task.verify(instance, workspace)  — domain-specific scoring
  │             5. Aggregate scores across instances
  │
  ├─ reflect: contrastive analysis (top-K vs bottom-K candidates)
  │     ├─ Structural reflection — which knobs/operators correlate with success
  │     └─ LLM reflection — domain-specific diagnosis + typed MutationSuggestions
  │
  └─ evolve: mutate workflows guided by reflection
        ├─ typed_suggestions → select_guided_operator (direct enum matching)
        └─ prompt_improvements → mutate_prompt hint text
```

### Key design decisions

1. **Task never sees the Workflow.** No `run(workflow)` method. The Task provides domain hooks; the InnerLoop drives the WorkflowExecutor. Knobs flow outer loop → workflow → executor, never through the Task.

2. **WorkflowExecutor receives domain context via `initial_context`.** The Task's `prompt(instance)` is seeded into `node_context[start_node]` before execution, so the first agent in the pipeline sees the domain-specific instruction.

3. **LLM reflection produces typed mutation suggestions.** The reflector sends one `claude -p` call per generation with contrastive data (top-K vs bottom-K). The LLM returns `{operator, target, rationale, value}` objects that `select_guided_operator` consumes via direct enum matching — no substring parsing.

4. **`validate_composition` respects domain tasks.** `required_capabilities` defaults to `None` (infer from scoring method). Setting `required_capabilities = []` explicitly declares "no software-dev capabilities needed" — domain tasks like chess-evolve pass validation with any workflow topology.

5. **Top-down budget for LLM context.** `_collect_individual_details` accepts a `char_budget` derived from `_LLM_PAYLOAD_BUDGET / num_individuals`. Experiment hypotheses (rich domain data) fill remaining budget after eval_details and instance_results.

### LLM contrastive reflection (Part A)
- Added `_llm_reflect()` to `OuterLoopReflector` — two-tier reflection (structural + LLM)
- Populates the previously dead `prompt_improvements` field via domain-agnostic contrastive prompt
- LLM now also returns typed `mutation_suggestions` with operator/target/rationale/value
- Experiment hypotheses surfaced via top-down budget allocation

### Typed mutation suggestions (Part B)
- `MutationSuggestion` dataclass on `ReflectionReport`
- `select_guided_operator` uses direct enum matching (was substring matching)
- `_parse_knob_suggestion` accepts typed objects (was string parsing)
- Backward compat: string fields still populated, fallback path when `typed_suggestions` empty

### OptKnob propagation (Part C)
- `--seed-workflow` required CLI flag for `calibrate`
- Dynamic import of `module.path:callable`, handles Package (calls `.compile()`) or Workflow
- Knobs survive into seed population → `KNOB_MUTATE` (25% default weight) now functional

### Task/Workflow separation
- Removed `Task.run()` entirely — four-hook interface only
- `_step_with_task`: setup → WorkflowExecutor(initial_context) → verify
- `WorkflowExecutor.initial_context` seeds domain prompt into start node

### validate_composition fix
- `required_capabilities` default changed from `[]` to `None` (three-way sentinel)
- Domain tasks with `required_capabilities = []` skip software-dev capability inference

## Test plan
- [x] 83 reflector tests (LLM reflection, typed suggestions, experiment context, budget)
- [x] 54 outer loop CLI tests (--seed-workflow, calibrate, reflect, evolve)
- [x] 50 compose tests (capability validation, sentinel semantics)
- [x] 19 inner loop task tests (split path, executor errors, empty prompt)
- [x] E2E integration: reflector → select_guided_operator typed suggestions
- [x] Adversarial QA: 21 criteria verified (malformed JSON, empty values, fallback paths)
- [x] CI: all checks pass (lint, types, tests on 3.11/3.12/3.13, codecov)

https://claude.ai/code/session_0156mZLsFKZy5y2FXCpqtD8S

---

## Follow-up: CLI selection-pressure fix

Closes #1476

The CLI outer-loop path was fitness-blind in two places (both fixed entirely in `factory/cli/outer_loop.py`; `engine.py` and `population.py` untouched):

- **`_cmd_evaluate` never wrote scores back to `population/population.json`** — every individual kept `score=0.0`. New `_propagate_scores_to_population()` runs after `results/gen{N}.json` is written, matches each evaluated mode to its individual via the `evolve-gen{N}-{id[:8]}` naming convention, and writes back `score` and `cost_usd`. It no-ops when the population file is absent or empty.
- **`_cmd_evolve` had zero selection pressure** — it mutated `modes[:population_size]`, so the worst candidate got as many offspring as the best. New `_load_mode_scores()` reads `results/gen{N}.json` (tolerating a missing or malformed file) and new `_select_parent()` does k=2 tournament selection. `_cmd_evolve` now restricts the parent pool to the current generation's candidate modes (excluding `-eval-` mirrors) and draws a fresh parent per offspring slot, printing the parent mode and score.

Tests added to `tests/test_outer_loop/test_cli.py`: score propagation end-to-end (plus missing-population and unmatched-mode paths), score loading, a 400-trial tournament distribution check, and a 200-offspring `_cmd_evolve` run asserting the worst parent gets zero offspring and the best beats the middle two.

Verification: `ruff` and `mypy` clean on the changed module; `pytest tests/test_outer_loop/ --timeout=120` → 544 passed, 2 failed. The two failures (`test_e2e.py::TestE2EEvolution::test_evolution_improves_over_seed`, `test_engine.py::TestSwarmEngineRun::test_run_terminates_on_budget`) are pytest-timeout expiries in `SwarmEngine.run`, whose `PROMPT_MUTATE` mutations make real LLM calls — both pass at `--timeout=600`, and neither imports `factory.cli.outer_loop`.


colehurwitz:
@ceo-review

colehurwitz:
@lambdabaa @akashgit worked on outerloop this weekend. mostly on fixing the wiring (bunch of small wiring issues - https://github.com/akashgit/remote-factory/pull/1467 and https://github.com/akashgit/remote-factory/pull/1466), but also building out the task -> inner loop -> outerloop architecture. would appreciate your thoughts on this!

factory/workflow/executor.py
```diff
@@ -100,6 +101,17 @@ def __init__(
...
+        if initial_context is not None:
+            start_node = workflow.nodes.get(workflow.start_node)
+            if not isinstance(start_node, AgentNode):
+                log.warning(
+                    "initial_context_ignored",
+                    start_node=workflow.start_node,
+                    node_type=type(start_node).__name__ if start_node else "missing",
+                    reason="initial_context only applies to AgentNode start nodes",
+                )
+            self.node_context[workflow.start_node] = initial_context
```
lambdabaa: The `initial_context only applies to AgentNode start nodes` warning never actually skips the assignment — `self.node_context[workflow.start_node] = initial_context` is inside the `if initial_context is not None:` block but outside the `isinstance` guard, so it runs for every node type.

Two problems follow:

1. **The log is false.** `initial_context_ignored` is emitted but the value is written to `node_context` immediately after. Nothing is ignored.

2. **It is actually consumed by non-AgentNode start nodes.** `node_context` is read in three places, not just `_run_agent` (AgentNode, line 833): also the LLM loop (line ~876) and `_build_gate_prompt` (GateNode, line 948). So when the workflow starts at a `GateNode` or `LLMNode`, the "ignored" context is silently applied to that node's prompt.

Suggested fix — actually guard the assignment (or broaden the allowed types to match what reads `node_context`):

```python
if initial_context is not None:
    start_node = workflow.nodes.get(workflow.start_node)
    if isinstance(start_node, AgentNode):
        self.node_context[workflow.start_node] = initial_context
    else:
        log.warning(
            "initial_context_ignored",
            start_node=workflow.start_node,
            node_type=type(start_node).__name__ if start_node else "missing",
            reason="initial_context only applies to AgentNode start nodes",
        )
```

If `GateNode`/`LLMNode` start nodes are meant to be supported too, widen the `isinstance` check and update the reason string accordingly.

colehurwitz:
Re: @lambdabaa's review on `initial_context` guard — fixed in f1d638d4. The assignment is now properly guarded behind `isinstance(start_node, AgentNode)`. Non-AgentNode start nodes get the warning and the context is genuinely skipped. Added 4 tests covering FnNode, GateNode, AgentNode, and None cases.

colehurwitz:
### `ce77dce1` — fix: CLI evolve now adds offspring to `population.json`

Follow-up to the score-propagation fix in `ad16d1b7`. `_cmd_evolve` registered offspring mode files but never mirrored them into `population/population.json`, so the population never grew past generation 0 — `_propagate_scores_to_population` found no individual matching a gen 1+ mode name and silently dropped every score, leaving the archive, status and convergence checks looking at gen 0 forever.

## Changes

- `_cmd_evolve` loads the population up front (`None` when no `population.json` exists, so projects that never seeded one evolve modes exactly as before) and saves it once after the offspring loop.
- New `_make_offspring_individual` helper builds the `Individual` for each registered offspring:
  - id is derived from the mode name returned by `registry.register` (which truncates ids to 8 chars) so it always matches what `_mode_suffix` extracts during score propagation;
  - `workflow_data` uses `Workflow.to_dict()` — the JSON-safe form that round-trips through `Workflow.from_dict` (plain `model_dump()` drops the `_type` discriminator);
  - `compute_features` is wrapped in try/except, defaulting to `()`.
- Older generations are kept in the population for lineage tracking; the population grows each generation.
- Scope: `factory/cli/outer_loop.py` and `tests/test_outer_loop/test_cli.py` only — `engine.py`, `population.py` and `models.py` are untouched, and no function signature changed.

## Tests

Four new tests in `tests/test_outer_loop/test_cli.py::TestEvolveAddsOffspringToPopulation`: offspring persistence (`generation`, `parent_id`, `mutation_record`, `score=0.0`), score propagation reaching gen 1 offspring, population growth across two generations, and the no-population-file fallback. The three regression tests fail on the pre-fix tree (`assert 0 == 2`, `assert 2 == 5`) and pass after.

`ruff check` and `mypy` clean on both files. `pytest tests/test_outer_loop/` (excluding the two e2e files): 273 passed. Two tests timed out only under batch load and pass standalone with the fix applied — `test_reflect_uses_saved_results_and_cycle_summary` (40s, spawns a real reflector) and `test_engine.py::TestSwarmEngineRun::test_run_terminates_on_budget` (0.18s standalone); neither touches the changed code path.

colehurwitz:
@ceo-review

PR #1481, opened 2026-09-09
feat: end-to-end task→workflow→outer-loop creation pipeline (#1478)

colehurwitz:
Closes #1478

### Phase 1: Shared `resolve_task()` + `--task` CLI flag
- Added `resolve_task()` 3-step resolver in `factory/task.py`: TOML file → Python file → `module:Class` import string
- Added `_load_task_from_python_file()` helper using importlib introspection
- Added `--task` flag to `factory outer-loop calibrate` and `factory outer-loop evaluate` commands
- Wired task resolution through `SwarmConfig.set_task()` for outer-loop integration
- Updated `docs/outer-loop.md` with Task Discovery documentation
- 12 tests covering all resolution paths, error cases, and equivalence

### Phase 2: `task-setup` workflow mode (W₁₄)
- Added `task_setup_workflow()` graph definition: fork(researcher_domain, researcher_verification) → join → CEO gate → strategist → user gate → builder → FnNode(validate_task) → archivist (10 nodes, 12 edges)
- Registered in `_get_builtin_registry()` and `WORKFLOW_META`
- Added `"task-setup"` to `CEO_MODES` in `_helpers.py`
- Added mode detection line in `ceo.md` prompt
- SKILL.md auto-generated (gitignored, regenerated by skill_cache)

### Phase 3: `--task` flag for create mode + OptKnob auto-generation
- Added `_build_task_aware_directive()` in `_task_builder.py` using same directive-injection pattern as Plugin Package and Update Existing Mode
- Added `--task` argument to CEO parser in `_parser_groups.py`
- Threaded `task_ref` through `ceo.py` → `_ceo_helpers.py` → `_task_builder.py`
- Auto-generates OptKnob mechanical derivation table: AgentNode.role→model, timeout→threshold, prompt_template→prompt
- Added "task-setup" mode suffix handling
- 18 tests covering directive injection, OptKnob tables, resolution failures, and integration

### Phase 4: `ScoringContract.threshold` + domain-level OptKnobs
- Added `threshold: float | None = None` field to `ScoringContract`
- Updated `TaskDefinition.from_toml()` to parse threshold from TOML scoring section
- Extended `_build_task_aware_directive()` to emit domain-level OptKnob instructions based on threshold presence
- Tests for threshold parsing, default values, and domain knob generation

### Phase 5: DataNode OptKnob integration (rebased onto PR #1483)
- Rebased onto PR #1483's branch (`factory/run-3110c50a`) — inherits DataNode, main, and review fixes
[...]

PR #1483, opened 2026-09-09
feat: DataNode graph primitive — first-class data loading in workflows

### LINKED ISSUE #1482: feat: DataNode graph primitive — first-class data loading in workflows
## Summary

Workflows have no way to declare "load data and iterate it." The `Task` four-hook interface (`instances`, `setup`, `prompt`, `verify`) defines data + eval, but it lives entirely outside the graph — `compose()` wires it into `InnerLoop` as Python glue, invisible to the executor, skill exports, and outer loop mutations. We need a `DataNode` primitive that makes data loading a first-class element in the workflow DAG.

## The gap

1. **No graph node for data** — the executor doesn't know about tasks or data sources. `InnerLoop._step_with_task()` calls `task.instances()` in a side-channel loop, bypassing the DAG entirely.
2. **No data flow through edges** — `TaskInstance` doesn't flow through `reads`/`writes`. Agent prompts come from `prompt_template` strings, not upstream data items. `task.prompt(instance)` is called by `InnerLoop`, not the graph.
3. **`Workflow.task` is inert** — it's a `str | None` on the `Workflow` model (primitives.py:277) but the executor never resolves or uses it.
4. **Outer loop is blind to data** — since data loading isn't structural, the optimizer can't reason about it (e.g. where data enters, how it fans out).

### DataNode primitive

```python
class DataNode(Node):
    """Node that loads data items and drives per-item execution of downstream subgraph."""
    
    model_config = ConfigDict(strict=True, extra="forbid")

    # Source: exactly one of these should be set
    task_ref: str = ""          # "module.path:ClassName" → Task.instances()
    source_path: str = ""       # glob, directory, file path  
    source_format: str = "auto" # auto, jsonl, csv, directory, glob
    inline_items: list[dict[str, Any]] = Field(default_factory=list)
    
    # Iteration control
    split: str = ""             # optional: train/val/test filter
    shuffle: bool = False
    limit: int = 0              # 0 = all items
```

### Data sources (without a Task)

The `DataNode` should work with or without a `Task`:

| Source | How it works |
|--------|-------------|
| `task_ref` | Resolves via `TaskRef`, calls `task.instances()` → yields `DataItem`s. Also wires `task.setup()` and `task.prompt()` into downstream nodes |
| `source_path` (directory) | Each subdirectory becomes a `DataItem(id=dirname, path=subdir)` |
| `source_path` (glob) | Each matched file becomes a `DataItem(id=filename, path=file)` |
| `source_path` (JSONL) | Each line becomes a `DataItem(id=line_number, input=parsed_json)` |
| `source_path` (CSV) | Each row becomes a `DataItem(id=row_index, input=row_dict)` |
| `inline_items` | Each dict becomes a `DataItem` directly — useful for small/test datasets |

### DataItem — standardized data object

```python
class DataItem(BaseModel):
    """Standardized data object flowing through the graph."""
    
    model_config = ConfigDict(strict=True, extra="forbid")

    id: str
    path: Path | None = None
    input: dict[str, Any] = Field(default_factory=dict)
    metadata: dict[str, Any] = Field(default_factory=dict)
    prompt: str = ""  # pre-computed prompt (from Task.prompt() or template)
```

`TaskInstance` maps directly: `TaskInstance(id, path, metadata)` → `DataItem(id, path, metadata=metadata, prompt=task.prompt(instance))`.

### Executor behavior

When the executor reaches a `DataNode`:
1. Resolve the data source → list of `DataItem`s
2. For each item, execute the downstream subgraph (everything between `DataNode` and the next `JoinNode` or terminal)
3. Write the current `DataItem` to the artifact the `DataNode.writes` declares, so downstream `AgentNode`s can read it via their `reads`
4. Aggregate results after all items complete

This is analogous to PyTorch's `DataLoader` driving a training loop — the `DataNode` is the iteration boundary.

### Relationship to existing pieces

- **`compose()`** stays as the pre-flight compatibility check. It validates that a `DataNode`-bearing workflow can handle its task's required capabilities.
- **`InnerLoop._step_with_task()`** delegates to the executor instead of side-channeling. The executor handles per-item iteration because it has the `DataNode`.
- **`Workflow.task`** can be deprecated in favor of `DataNode.task_ref` — or kept as sugar that the executor resolves to find the `DataNode`.
- **`GateNode`** remains the in-graph eval checkpoint. A dedicated `EvalNode` (wrapping `Task.verify()`) can be added later if the outer loop needs eval as a mutable graph element.

### Outer loop integration

`DataNode` becomes visible to:
- `compute_features()` in `similarity.py` — new feature dimension (has_data_node, data_source_type)
- Mutations — `DataNode` should typically be frozen (you mutate the workflow around the data, not the data itself), add to `frozen_node_ids` by default
- MAP-Elites — data loading strategy could become an archive dimension

## Files to modify

- `factory/workflow/primitives.py` — add `DataNode`, `DataItem`, update `NodeType` union and `from_dict`
- `factory/workflow/executor.py` — handle `DataNode` iteration in the execution loop
- `factory/workflow/validation.py` — validate `DataNode` constraints (exactly one source set, downstream subgraph structure)
- `factory/compose.py` — update `ModeCapabilities.from_workflow()` to recognize `DataNode`
- `factory/inner_loop.py` — simplify `_step_with_task()` to delegate to executor when workflow has a `DataNode`
- `factory/workflow/definitions.py` — update workflow definitions that use tasks to include a `DataNode`
- `factory/outer_loop/similarity.py` — add `DataNode` to feature computation
- `factory/outer_loop/mutations.py` — auto-freeze `DataNode` by default
- Tests for all of the above

## Non-goals (for now)

- `EvalNode` — eval works fine through `Task.verify()` + `GateNode`. Add later if needed.
- Streaming/lazy iteration — start with materialized list, optimize later.
- Cross-workflow data sharing — each workflow owns its data source.

## Related

- #1478 — cost_usd always 0.00 (separate bug in cost propagation, not data loading)
- #1479 — end-to-end task→workflow→outer-loop pipeline (DataNode is the missing link)

colehurwitz:
Closes #1482

## What and why

Workflows have no way to declare "load data and iterate it." Today, per-instance iteration lives as a side-channel Python `for` loop in `InnerLoop._step_with_task()` — invisible to the executor, skill exports, and the outer loop's evolutionary search. This PR adds `DataNode`, a first-class graph primitive that makes data loading a structural element of the workflow DAG.

**Before:** InnerLoop calls `task.instances()` in a manual loop the executor can't see:
```
InnerLoop (Python for-loop, invisible to executor):
    for problem in task.instances():
        executor.execute(workflow)  ← executor has no idea it's in a loop
```

**After:** The executor owns the iteration via DataNode:
```
DataNode (loads items, visible in graph)
    └─ subgraph: study → build → qa → eval  (full workflow per item)
→ JoinNode (aggregate)
```

This means: the executor can parallelize items (`Semaphore`-throttled), fault-isolate per item (one failure → score 0.0, others continue), render iteration in SKILL.md playbooks, and expose data-loading to the outer loop's MAP-Elites search.

## How it works

`DataNode` has three data sources — use exactly one:

| Source | What it does |
|--------|-------------|
| `inline_items` | Hardcoded DataItems in the workflow definition |
| `source_path` | Points at a directory, JSONL, or CSV file |
| `task_ref` | Resolves a Task class, calls `task.instances()` → DataItems |

When the executor reaches a DataNode, it:
1. Resolves the source → `list[DataItem]`
2. Applies `split` / `shuffle` / `limit` filters
3. Validates item count against `max_items` safety ceiling (default 500)
4. Extracts the subgraph (`subgraph_entry` → `subgraph_exit`)
5. Runs the subgraph once per item, throttled by `asyncio.Semaphore(parallelism)`
6. Each item is fault-isolated: exception → `{score: 0.0, error: ...}`, doesn't halt others
7. Aggregates per-item results as JSON at the DataNode level
8. Writes `current_item.json` before each subgraph run for item visibility (cleaned up after)

DataNode works **anywhere the executor runs** — headless mode, skill export, InnerLoop delegation, standalone. It doesn't require InnerLoop or a Task; `inline_items` and `source_path` work without either.

## Architecture

DataNode is modeled on `SubgraphForkNode` (runtime-resolved cardinality), not `ForkNode` (static targets). It shares the same `subgraph_entry` / `subgraph_exit` / `parallelism` shape and reuses `_collect_subgraph_nodes()` + `Workflow.subgraph()`.

```
                    ┌─────────────────────────────────────────┐
                    │              DataNode                    │
                    │                                         │
                    │  source: task_ref / source_path / inline │
                    │  filters: split, shuffle, limit          │
                    │  safety: max_items (default 500)         │
                    │  concurrency: parallelism (default 1)    │
                    │                                         │
                    │  ┌─────────────────────────────────────┐ │
                    │  │  subgraph (per item)                │ │
                    │  │  entry → node → ... → exit          │ │
                    │  └─────────────────────────────────────┘ │
                    └─────────────────────────────────────────┘
                                      │
                                      ▼
                                  next node
```

## Changes by layer

**Layer 2 — Workflow engine (the core):**
- `primitives.py` — `DataItem(BaseModel)` + `DataNode(Node)` with Verdict-style `model_validator` for exactly-one-source, `NodeType` union extended, `_NODE_TYPE_MAP` updated
- `executor.py` — `_execute_data()` with Semaphore-throttled concurrent per-item execution, per-item fault isolation, per-item worktree isolation (parallelism>1), current_item.json contract, sub-executor inherits parent's `completed_files` so subgraph reads don't timeout
- `validation.py` — Both `_validate_fork_join_nodes()` and `_validate_reachability()` updated + `_validate_datanode_edges()` rejects explicit DataNode→subgraph edges
- `skill_export.py` — DataNode rendering branch + subgraph skip-set + trailing `else` warning for unmatched types

**Layer 2b — Outer loop:**
- `similarity.py` — Appended `has_data_node` feature axis (9-tuple, not inserted — existing indices stable)
- `population.py` — `diversity_metric()` generalized from hardcoded `range(4)` to dynamic axis count
- `engine.py` — `_auto_frozen_nodes()` auto-includes DataNode IDs in `frozen_node_ids` so mutations can't remove them

**Layer 3 — InnerLoop glue:**
- `inner_loop.py` — Cached `_workflow_has_data_node()` check, `_step_with_data_node()` delegation with direct DataNode ID score lookup + `instance_results` populated on CycleRecord
- `compose.py` — `DataNode` → `CAN_ITERATE` capability; DataNode workflows skip build-pipeline capability requirements (HAS_BUILDER, CAN_RUN_TESTS, CAN_MODIFY_CODE)

## What's NOT in this PR

- `task.verify()` inside the graph (deferred `EvalNode` concept) — verify stays in InnerLoop for now
- `--mode create` auto-inserting DataNode at graph root
- Migrating existing workflows to DataNode (nothing to migrate — `Workflow.task` is unused)
- Streaming/lazy iteration — materialized list for now
- TaskInstance/DataItem unification — mapping layer is 3 lines, revisit if it becomes friction

### Spec deviations from issue #1482

- `source_format` is `Literal['directory'|'jsonl'|'csv']`, not `'auto'` with auto-detection (simpler, explicit)
- No glob support in `source_path` (direct path only)
- `DataItem.input` became `DataItem.metadata` (dict instead of string)
- `task_ref` wiring done per-item in executor, not via node params (reviewer agreed this is better)
- `parallelism` default changed from 3 to 1 (safe default; `parallelism>1` uses per-item worktrees for isolation)
- `shuffle_seed` field added for reproducibility (unseeded shuffle derives seed from `node_id:run_id`)
- `current_item.json` contract added for subgraph item visibility (written before each subgraph run, cleaned up after)

## Test plan

- [x] 55 DataNode-specific tests (`tests/test_data_node.py`) — models, executor, fault isolation, source_path (directory/jsonl/csv), validation, skill export, features, completed_files inheritance, parallelism default, seeded shuffle, JSONL error isolation, current_item.json, direct score lookup, instance_results, compose validation
- [x] 6 integration tests (`tests/test_inner_outer_loop.py`) — full Task→InnerLoop→Executor→outer loop pipeline + compose DataNode workflow
- [x] 3 mutation tests (`tests/test_outer_loop/test_mutations.py`) — auto-freeze DataNode IDs
- [x] 19/19 `test_inner_loop_task.py` pass unmodified (backward compat)
- [x] 41/41 outer loop tests pass (similarity + population assertions updated)
- [x] `ruff check` clean

**Note:** Existing outer-loop checkpoints (`grid.json`/`population.json`) are incompatible across this change due to the 9-tuple feature vector. Mid-flight outer-loop runs should not resume across it.

### DIFF
--- a/factory/inner_loop.py
+++ b/factory/inner_loop.py
@@ -388,6 +388,7 @@ class InnerLoop:
 
         t0 = time.monotonic()
 
+        assert self.workflow is not None
         executor = WorkflowExecutor(
             self.workflow,
             self.project_dir,
--- a/factory/inner_loop.py
+++ b/factory/inner_loop.py
@@ -383,6 +383,7 @@ class InnerLoop:
     def _step_with_data_node(self, directives: dict[str, Any] | None = None) -> CycleRecord:
         """Delegate to the executor when the workflow contains a DataNode."""
         import asyncio
+        import json
 
         from factory.workflow.executor import WorkflowExecutor
 
@@ -397,6 +398,17 @@ class InnerLoop:
 
         duration_s = time.monotonic() - t0
         score = 1.0 if exec_result.success else 0.0
+
+        # Aggregate per-item verify scores from DataNode output if available
+        for _nid, output in exec_result.node_outputs.items():
+            try:
+                parsed = json.loads(output)
+                if isinstance(parsed, list) and parsed and "score" in parsed[0]:
+                    scores = [r["score"] for r in parsed]
+                    score = sum(scores) / len(scores) if scores else 0.0
+                    break
+            except (json.JSONDecodeError, TypeError, KeyError):
+                continue
 
         record = CycleRecord(
             cycle_number=self._step_count + 1,
--- a/factory/inner_loop.py
+++ b/factory/inner_loop.py
@@ -27,7 +27,7 @@ from pathlib import Path
 from typing import Any, Protocol, runtime_checkable
 
 from factory.cycle_analyzer import CycleAnalyzer, CycleRecord
-from factory.workflow.primitives import Workflow
+from factory.workflow.primitives import DataNode, Workflow
 
 
 @dataclass
@@ -135,6 +135,7 @@ class InnerLoop:
         self.instance = instance
         self._step_count = 0
         self._history: list[CycleRecord] = []
+        self._has_data_node: bool | None = None
         self._validate_frozen_nodes()
 
         # When task is set, derive flat fields from it for backward compat
@@ -256,6 +257,14 @@ class InnerLoop:
         self._history.append(record)
         return record
 
+    def _workflow_has_data_node(self) -> bool:
+        if self._has_data_node is None:
+            self._has_data_node = (
+                self.workflow is not None
+                and any(isinstance(n, DataNode) for n in self.workflow.nodes.values())
+            )
+        return self._has_data_node
+
     def _step_with_task(self, directives: dict[str, Any] | None = None) -> CycleRecord:
         """Task-driven step: setup → WorkflowExecutor → verify per instance."""
         assert self.task is not None
@@ -268,6 +277,9 @@ class InnerLoop:
 
         if directives:
             self._write_directives(directives)
+
+        if self._workflow_has_data_node():
+            return self._step_with_data_node(directives)
 
         t0 = time.monotonic()
         workflow = self.workflow
@@ -362,6 +374,49 @@ class InnerLoop:
             experiments=0,
             test_score=aggregate_score,
             instance_results=instance_results,
+        )
+
+        self._step_count += 1
+        self._history.append(record)
+        return record
+
+    def _step_with_data_node(self, directives: dict[str, Any] | None = None) -> CycleRecord:
+        """Delegate to the executor when the workflow contains a DataNode."""
+        import asyncio
+
+        from factory.workflow.executor import WorkflowExecutor
+
+        t0 = time.monotonic()
+
+        executor = WorkflowExecutor(
+            self.workflow,
+            self.project_dir,
+        )
+        exec_result = asyncio.run(executor.execute())
+
+        duration_s = time.monotonic() - t0
+        score = 1.0 if exec_result.success else 0.0
+
+        record = CycleRecord(
+            cycle_number=self._step_count + 1,
+            mode=self.mode,
+            started_at=None,
+            ended_at=None,
+            duration_s=duration_s,
+            score_start=None,
+            score_end=score,
+            score_delta=None,
+        )
+        record.frozen_nodes = sorted(self.frozen_nodes)
+        record.mutable_node_ids = sorted(self.mutable_nodes())
+
+        self._write_cycle_summary(
+            returncode=0 if exec_result.success else 1,
+            event_offset=0,
+            duration_ms=int(duration_s * 1000),
+            builder_committed=False,
+            experiments=0,
+            test_score=score,
         )
 
         self._step_count += 1
--- a/factory/inner_loop.py
+++ b/factory/inner_loop.py
@@ -411,17 +411,33 @@ class InnerLoop:
 
         duration_s = time.monotonic() - t0
         score = 1.0 if exec_result.success else 0.0
+        instance_results: list[dict[str, Any]] | None = None
 
-        # Aggregate per-item verify scores from DataNode output if available
-        for _nid, output in exec_result.node_outputs.items():
+        # Direct lookup: find DataNode ID and read its output
+        data_node_id: str | None = None
+        for nid, n in self.workflow.nodes.items():
+            if isinstance(n, DataNode):
+                data_node_id = nid
+                break
+
+        if data_node_id is not None and data_node_id in exec_result.node_outputs:
             try:
-                parsed = json.loads(output)
-                if isinstance(parsed, list) and parsed and "score" in parsed[0]:
-                    scores = [r["score"] for r in parsed]
-                    score = sum(scores) / len(scores) if scores else 0.0
-                    break
+                parsed = json.loads(exec_result.node_outputs[data_node_id])
+                if isinstance(parsed, list) and parsed:
+                    scores = [r["score"] for r in parsed if "score" in r]
+                    if scores:
+                        score = sum(scores) / len(scores)
+                    instance_results = [
+                        {
+                            "instance_id": item.get("item_id", ""),
+                            "score": item.get("score", 0.0),
+                            "passed": item.get("passed", False),
+                        }
+                        for item in parsed
+                        if isinstance(item, dict)
+                    ]
             except (json.JSONDecodeError, TypeError, KeyError):
-                continue
+                pass
 
         record = CycleRecord(
             cycle_number=self._step_count + 1,
@@ -432,6 +448,7 @@ class InnerLoop:
             score_start=None,
             score_end=score,
             score_delta=None,
+            instance_results=instance_results,
         )
         record.frozen_nodes = sorted(self.frozen_nodes)
         record.mutable_node_ids = sorted(self.mutable_nodes())
@@ -443,6 +460,7 @@ class InnerLoop:
             builder_committed=False,
             experiments=0,
             test_score=score,
+            instance_results=instance_results,
         )
 
         self._step_count += 1
--- a/factory/outer_loop/population.py
+++ b/factory/outer_loop/population.py
@@ -94,6 +94,12 @@ class Population:
         return pop
 
 
+# Structural axes from compute_features() 9-tuple:
+# 0=depth, 1=fork_degree, 2=agent_count, 3=gate_count, 8=has_data_node
+# Indices 4-7 are hash buckets (edge_sig, param, prompt, knob) — excluded.
+STRUCTURAL_AXES: tuple[int, ...] = (0, 1, 2, 3, 8)
+
+
 class MAPElitesArchive:
     """4D fixed-resolution grid archive for quality-diversity search.
 
@@ -193,20 +199,22 @@ class MAPElitesArchive:
         return front
 
     def diversity_metric(self) -> float:
-        """Fraction of occupied cells relative to a reasonable grid size estimate.
+        """Fraction of occupied cells in the structural-axis subspace.
 
         Returns 0.0 for empty archive, approaches 1.0 as more cells are filled.
-        Uses the first 5 structural axes (depth, fork_degree, agent_count,
-        gate_count, has_data_node) for diversity estimation.
+        Uses 5 structural axes from the 9-tuple returned by compute_features():
+          depth (0), fork_degree (1), agent_count (2), gate_count (3), has_data_node (8).
+        Hash-bucket axes (4-7: edge_sig, param, prompt, knob) are excluded because
+        their high cardinality inflates the denominator without reflecting structural diversity.
         """
         if not self._grid:
             return 0.0
         sample_key = next(iter(self._grid))
-        n_axes = min(len(sample_key), 5)
-        unique_per_axis: list[set[int]] = [set() for _ in range(n_axes)]
+        axes = [a for a in STRUCTURAL_AXES if a < len(sample_key)]
+        unique_per_axis: list[set[int]] = [set() for _ in axes]
         for key in self._grid:
-            for i in range(n_axes):
-                unique_per_axis[i].add(key[i])
+            for idx, a in enumerate(axes):
+                unique_per_axis[idx].add(key[a])
         total_possible = 1
         for s in unique_per_axis:
             total_possible *= max(len(s), 1)
--- a/factory/outer_loop/population.py
+++ b/factory/outer_loop/population.py
@@ -196,14 +196,17 @@ class MAPElitesArchive:
         """Fraction of occupied cells relative to a reasonable grid size estimate.
 
         Returns 0.0 for empty archive, approaches 1.0 as more cells are filled.
+        Uses the first 5 structural axes (depth, fork_degree, agent_count,
+        gate_count, has_data_node) for diversity estimation.
         """
         if not self._grid:
             return 0.0
-        unique_per_axis: list[set[int]] = [set() for _ in range(4)]
+        sample_key = next(iter(self._grid))
+        n_axes = min(len(sample_key), 5)
+        unique_per_axis: list[set[int]] = [set() for _ in range(n_axes)]
         for key in self._grid:
-            for i, v in enumerate(key):
-                if i < 4:
-                    unique_per_axis[i].add(v)
+            for i in range(n_axes):
+                unique_per_axis[i].add(key[i])
         total_possible = 1
         for s in unique_per_axis:
             total_possible *= max(len(s), 1)
--- a/factory/outer_loop/population.py
+++ b/factory/outer_loop/population.py
@@ -218,7 +218,8 @@ class MAPElitesArchive:
         total_possible = 1
         for s in unique_per_axis:
             total_possible *= max(len(s), 1)
-        return len(self._grid) / max(total_possible, 1)
+        structural_cells = {tuple(key[a] for a in axes) for key in self._grid}
+        return len(structural_cells) / max(total_possible, 1)
 
     def save(self, directory: Path) -> None:
         """Serialize the archive to a directory."""
--- a/factory/workflow/executor.py
+++ b/factory/workflow/executor.py
@@ -636,6 +636,7 @@ class WorkflowExecutor:
 
     async def _execute_data(self, node_id: str, node: DataNode) -> None:
         """Execute a DataNode: resolve items, run subgraph per item with fault isolation."""
+        import hashlib as _hashlib
         import random as _random
         import subprocess as _sp
 
@@ -685,6 +686,17 @@ class WorkflowExecutor:
                 raise FileNotFoundError(
                     f"DataNode '{node_id}': source_path not found: {node.source_path}"
                 )
+            # Validate path kind matches source_format
+            if node.source_format == "directory" and not src.is_dir():
+                raise ValueError(
+                    f"DataNode '{node_id}': source_format='directory' requires a directory, "
+                    f"got a file at {node.source_path}"
+                )
+            if node.source_format in ("jsonl", "csv") and not src.is_file():
+                raise ValueError(
+                    f"DataNode '{node_id}': source_format='{node.source_format}' requires a file, "
+                    f"got a directory at {node.source_path}"
+                )
             if node.source_format == "directory" and src.is_dir():
                 for child in sorted(src.iterdir()):
                     if child.is_dir():
@@ -731,7 +743,10 @@ class WorkflowExecutor:
             seed = (
                 node.shuffle_seed
                 if node.shuffle_seed is not None
-                else hash(f"{node_id}:{self.run_id}") % (2**32)
+                else int.from_bytes(
+                    _hashlib.sha256(f"{node_id}:{self.run_id}".encode()).digest()[:8],
+                    "big",
+                )
             )
             _random.Random(seed).shuffle(task_instances)
         if node.limit is not None and node.limit > 0:
@@ -837,7 +852,7 @@ class WorkflowExecutor:
                             item_project_path,
                             agent_pool=self.agent_pool,
                             dry_run=self.dry_run,
-                            initial_context=item.prompt,
+                            initial_context=item.prompt or None,
                         )
                         item_executor.completed_files = self.completed_files | disk_reads
                         item_result = await item_executor.execute()
--- a/factory/workflow/executor.py
+++ b/factory/workflow/executor.py
@@ -637,6 +637,7 @@ class WorkflowExecutor:
     async def _execute_data(self, node_id: str, node: DataNode) -> None:
         """Execute a DataNode: resolve items, run subgraph per item with fault isolation."""
         import random as _random
+        import subprocess as _sp
 
         from factory.task import Task as _Task
         from factory.task import TaskInstance as _TaskInstance
@@ -679,17 +680,40 @@ class WorkflowExecutor:
             src = _Path(node.source_path)
             if not src.is_absolute():
                 src = self.project_path / src
+            # Raise if source_path doesn't exist
+            if not src.exists():
+                raise FileNotFoundError(
+                    f"DataNode '{node_id}': source_path not found: {node.source_path}"
+                )
             if node.source_format == "directory" and src.is_dir():
                 for child in sorted(src.iterdir()):
                     if child.is_dir():
                         task_instances.append((DataItem(id=child.name, path=str(child)), None))
             elif node.source_format == "jsonl" and src.is_file():
-                for idx, line in enumerate(src.read_text().splitlines()):
+                error_count = 0
+                lines = src.read_text().splitlines()
+                total_non_empty = 0
+                for idx, line in enumerate(lines):
                     if line.strip():
-                        task_instances.append((DataItem(
-                            id=str(idx),
-                            metadata=json.loads(line),
-                        ), None))
+                        total_non_empty += 1
+                        try:
+                            task_instances.append((DataItem(
+                                id=str(idx),
+                                metadata=json.loads(line),
+                            ), None))
+                        except json.JSONDecodeError:
+                            error_count += 1
+                            log.warning(
+                                "jsonl_parse_error",
+                                node_id=node_id,
+                                line_number=idx + 1,
+                                line_preview=line.strip()[:100],
+                            )
+                if error_count > 0 and error_count == total_non_empty:
+                    raise ValueError(
+                        f"DataNode '{node_id}': all {error_count} JSONL lines "
+                        f"failed to parse"
+                    )
             elif node.source_format == "csv" and src.is_file():
                 import csv
                 with src.open(newline="") as f:
@@ -704,9 +728,21 @@ class WorkflowExecutor:
                 if item.metadata.get("split") == node.split
             ]
         if node.shuffle:
-            _random.shuffle(task_instances)
+            seed = (
+                node.shuffle_seed
+                if node.shuffle_seed is not None
+                else hash(f"{node_id}:{self.run_id}") % (2**32)
+            )
+            _random.Random(seed).shuffle(task_instances)
         if node.limit is not None and node.limit > 0:
             task_instances = task_instances[:node.limit]
+
+        if len(task_instances) == 0:
+            log.warning(
+                "data_source_empty",
+                node_id=node_id,
+                source=str(node.source_path or node.task_ref or "inline"),
+            )
 
         if len(task_instances) > node.max_items:
             raise ValueError(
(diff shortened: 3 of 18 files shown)


colehurwitz:
@lambdabaa This is the new datanode primitive. It should handle IO into workflows including tasks and non-task paths.

colehurwitz:
Waiting for this to land before finishing https://github.com/akashgit/remote-factory/pull/1481

colehurwitz:
@ceo-review

colehurwitz:
@ceo-review

factory/workflow/executor.py
```diff
@@ -628,6 +634,182 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+                    if resolved_task is not None and inst is not None:
+                        resolved_task.setup(inst, self.project_path)
+                        item = DataItem(
+                            id=inst.id,
+                            path=str(inst.path) if inst.path else None,
+                            metadata=inst.metadata,
+                            prompt=resolved_task.prompt(inst),
+                        )
+
+                    item_executor = WorkflowExecutor(
```
lambdabaa: **Items share one workspace (blocking)**

Every item sub-executor runs against `self.project_path`, with default `parallelism=3`. So three items run concurrently in the same directory. The manual loop this replaces ran instances sequentially (slow but safe), and SubgraphForkNode, the node this is modeled on, gives each branch its own worktree. Here, concurrent agents will race on the same files: `builder-latest.md`, strategy state, whatever the subgraph writes.

Two ways out, either works for me:
- per-item isolation (a worktree per item, reusing the SubgraphFork machinery, or a per-item subdirectory contract)
- default `parallelism=1`, with concurrency as explicit opt-in documented as requiring isolation

Also note `task.setup(inst, self.project_path)` runs concurrently against the same workspace for task_ref nodes.

factory/workflow/executor.py
```diff
@@ -628,6 +634,182 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+                        sub_workflow.model_copy(deep=True),
+                        self.project_path,
+                        agent_pool=self.agent_pool,
+                        dry_run=self.dry_run,
+                        initial_context=item.prompt,
```
lambdabaa: **The subgraph can't see the item (blocking)**

`initial_context=item.prompt` is the only channel from item to subgraph, and it has two holes:

1. If `subgraph_entry` is an FnNode, the executor logs `initial_context_ignored` and drops the prompt entirely (I hit this while testing).
2. For `source_path` items (jsonl/csv/directory), `prompt` is always `""`. The parsed row lives in `metadata` and never reaches the subgraph at all. Every item runs an identical, context-free subgraph.

Right now that makes iteration over source_path data structurally impressive but functionally a no-op. Needs a real channel: write the current item to a well-known file before each run (something like `.factory/current_item.json`), or template item fields into node params.

factory/workflow/executor.py
```diff
@@ -628,6 +634,182 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+            ),
+        )
+
+        start = time.monotonic()
+
+        # Resolve data items from exactly one source, keeping TaskInstances for task_ref
+        task_instances: list[tuple[DataItem, _TaskInstance | None]] = []
+        resolved_task: _Task | None = None
+
+        if node.inline_items:
```
lambdabaa, line 664: **Explicit edge to subgraph_entry double-executes (blocking)**

The natural way to author this is to draw an edge from the DataNode to its subgraph, which the PR's own `test_from_dict_roundtrip` does. I tried it: `nodes_executed` came back as 3 for a 1-item workflow, because `_execute_data` runs the subgraph per item, then `_next_unconditional` follows the explicit edge and runs the subgraph again in the parent, outside item context.

I'd have validation reject edges from the DataNode into its own subgraph (or from subgraph nodes back out), since this will be the first thing users try. Alternatively the executor could skip subgraph nodes when continuing past the DataNode.

factory/workflow/executor.py
```diff
@@ -628,6 +634,182 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+            for inst in resolved_task.instances():
+                task_instances.append((
+                    DataItem(
+                        id=inst.id,
+                        path=str(inst.path) if inst.path else None,
+                        metadata=inst.metadata,
+                    ),
+                    inst,
+                ))
+        elif node.source_path:
```
lambdabaa, line 679: **Empty or missing data silently succeeds (blocking)**

`test_nonexistent_path_yields_empty` codifies: typo'd `source_path` results in 0 items and `success=True`. Same for format mismatches like directory format pointing at a file. A data pipeline that evaluates nothing and passes is worse than one that fails loudly, especially under the outer loop where an empty eval scores 0.0 for reasons that have nothing to do with workflow quality.

I'd raise on a nonexistent path, and at minimum warn (event + log) when a resolved source yields zero items. Related: a malformed JSONL line raises inside `_execute_data` at node level rather than being item-isolated, which cuts against the fault isolation story.

factory/workflow/executor.py
```diff
@@ -628,6 +634,182 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+                        task_instances.append((DataItem(id=str(idx), metadata=dict(row)), None))
+
+        # Apply split/shuffle/limit filters to the paired list
+        if node.split != "all":
+            task_instances = [
+                (item, inst) for item, inst in task_instances
+                if item.metadata.get("split") == node.split
+            ]
+        if node.shuffle:
+            _random.shuffle(task_instances)
```
lambdabaa: **Unseeded shuffle breaks reproducibility (blocking, small fix)**

`_random.shuffle` with no seed, while `FitnessCache` dedups candidates by structural hash. Same candidate, same hash, cache hit, but a different shuffle would have produced different scores on any re-evaluation path (holdout, resume, cache invalidation). Today the cache masks it. That makes those paths irreproducible.

Seed it from the config or run_id.

factory/inner_loop.py
```diff
@@ -381,6 +393,62 @@ def _step_with_task(self, directives: dict[str, Any] | None = None) -> CycleReco
...
+            self.workflow,
+            self.project_dir,
+        )
+        exec_result = asyncio.run(executor.execute())
+
+        duration_s = time.monotonic() - t0
+        score = 1.0 if exec_result.success else 0.0
+
+        # Aggregate per-item verify scores from DataNode output if available
+        for _nid, output in exec_result.node_outputs.items():
```
lambdabaa: **Score sniffing is fragile (non-blocking)**

This walks every node output, json-parses it, and takes the first that looks like a list with a score key. Two nits: `"score" in parsed[0]` is a substring check if the element happens to be a string, and any node whose output is a JSON list of dicts with a score key could get picked by accident.

The workflow knows which node is the DataNode, so reading that node's output directly would be simpler and safer.

Related: this path also drops `instance_results` from both the CycleRecord and `_write_cycle_summary`. The manual path populates both, so downstream consumers (CycleAnalyzer, SwarmEvaluator's CycleRecordCache) lose all per-instance detail. The data is right there in `exec_result.node_outputs`.

factory/workflow/primitives.py
```diff
@@ -191,6 +191,50 @@ class SubgraphForkNode(Node):
...
+
+
+class DataNode(Node):
+    """Node that loads data items and drives per-item execution of a subgraph."""
+
+    model_config = ConfigDict(strict=True, extra="forbid")
+
+    task_ref: str | None = None
+    source_path: str | None = None
+    source_format: Literal["directory", "jsonl", "csv"] | None = None
```
lambdabaa, line 214: **Spec deviations from #1482 worth documenting (non-blocking)**

A few places where the implementation diverges from the issue, all of which look like reasonable simplifications to me:
- the issue specifies `source_format: "auto"` with auto-detection and glob support; here it's a Literal of three formats, required, no auto, no glob
- the issue's `DataItem.input` field became `metadata`
- the issue says task_ref wires setup/prompt into downstream nodes; the implementation does it per-item inside the executor (I actually think that's better)

Since the PR body still echoes the issue's table, a reader can't tell what shipped. A line or two in the "What's NOT in this PR" section would cover it.

lambdabaa (changes requested):
Reviewed this in depth, including running the branch locally (all 207 new/modified tests pass in 7s, nice) and poking at the executor behavior with small scripts. First, the good stuff: the direction is right. Making data loading visible to the executor, skill export, and the outer loop is the correct move, and modeling on SubgraphForkNode was the right call. The scoping section in the description is honest and the completed_files inheritance fix shows you actually debugged the integration, not just the happy path.

That said, I'm requesting changes. The first three comments below feel like the same root cause to me: the contract between a DataNode, its items, and the subgraph isn't pinned down yet. Until it is, I don't think workflows authored against this are safe to run:

1. Items run concurrently in one shared workspace, so they clobber each other
2. The subgraph can't actually see the item it's running for (unless it starts with an AgentNode and the item has a prompt)
3. Drawing the natural edge from the DataNode to its subgraph runs everything twice
4. A typo'd source_path silently succeeds with zero items
5. shuffle is unseeded, which breaks reproducibility for the outer loop

I verified each of these by running code against the branch, so the line comments have concrete repro notes. Items 6 and 7 are smaller and could be follow-up issues if you'd rather keep the PR moving.

Also worth noting the Sentrux report flagged _execute_data at cc=34 (repo max is 30). I'd split item resolution, filtering, and execution into separate methods. That would also make the fault paths easier to test individually, which is where most of my comments live.

Happy to pair on any of this or re-review quickly once it's updated.

factory/workflow/executor.py
colehurwitz: Fixed in fc3508a8. Default parallelism=1 (safe), parallelism>1 creates per-item git worktrees (reuses SubgraphForkNode machinery). task.setup() uses item worktree path.

factory/workflow/executor.py
colehurwitz: Fixed in fc3508a8. Before each subgraph run, writes `item.model_dump()` to `{item_project_path}/.factory/current_item.json`. This covers FnNode entries (where initial_context was dropped) and source_path items (where prompt was always empty). Cleanup in `finally` block. With parallelism>1, each item has its own worktree (fix 1) so no race on the file.

factory/workflow/executor.py
colehurwitz: Fixed in fc3508a8. Added `_validate_datanode_edges()` to validation.py — BFS collects subgraph node set via `_collect_subgraph_nodes()`, then checks all edges. If source is a DataNode and target is in its subgraph, validation rejects it with a descriptive error about double-execution. Called from `validate_workflow()`.

factory/workflow/executor.py
colehurwitz: Fixed in fc3508a8. Three tiers:
1. **Nonexistent source_path** → `FileNotFoundError` (fail fast)
2. **Malformed JSONL lines** → per-line `try/except JSONDecodeError`, bad lines skipped with warning (line number + preview). If ALL lines fail → `ValueError`. Good lines preserved.
3. **Zero items after filtering** → `data_source_empty` warning via structlog

factory/workflow/executor.py
colehurwitz: Fixed in fc3508a8. Added `shuffle_seed: int | None` field to DataNode. Uses `random.Random(seed).shuffle()` — isolated RNG, no global state. Fallback seed derived from `hash(f'{node_id}:{run_id}') % 2**32`. Note: fallback uses Python's PYTHONHASHSEED-randomized `hash()` so it's deterministic within a process but not across restarts. For true cross-process reproducibility, set `shuffle_seed` explicitly. Could improve to `hashlib` in a follow-up.

factory/inner_loop.py
colehurwitz: Fixed in fc3508a8. Direct lookup by DataNode ID (`isinstance(n, DataNode)`) instead of walking all outputs. Also populates `instance_results` on both `CycleRecord` and `_write_cycle_summary` — downstream consumers (CycleAnalyzer, SwarmEvaluator CycleRecordCache) now get per-instance detail.

factory/workflow/primitives.py
colehurwitz: Done in fc3508a8. Added "Spec deviations from issue #1482" section to the PR body. Documents: source_format is Literal not auto, no glob, input→metadata rename, task_ref wiring in executor, parallelism default 3→1, shuffle_seed added, current_item.json contract.

Also included the compose() fix for DataNode workflows (issue #1488 Gap 0) in this commit — skips HAS_BUILDER/CAN_RUN_TESTS/CAN_MODIFY_CODE capability checks when workflow has a DataNode.

colehurwitz:
@ceo-review

colehurwitz:
@lambdabaa i think this is ready for another review :)

colehurwitz:
@ceo-review

factory/workflow/executor.py
```diff
@@ -628,6 +634,285 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+            from pathlib import Path as _Path
+            src = _Path(node.source_path)
+            if not src.is_absolute():
+                src = self.project_path / src
+            # Raise if source_path doesn't exist
+            if not src.exists():
+                raise FileNotFoundError(
+                    f"DataNode '{node_id}': source_path not found: {node.source_path}"
+                )
+            if node.source_format == "directory" and src.is_dir():
```
lambdabaa, line 700: If the path exists but is the wrong kind for the format (`directory` format on a file, `jsonl`/`csv` on a directory), these guards silently fall through to zero items, and the run succeeds with a `data_source_empty` warning. I verified: `source_format='directory'` pointed at a regular file gives `success=True` with zero items and a warning that says "empty", not "wrong format".

Since a nonexistent path already raises `FileNotFoundError`, the wrong-kind case should raise too — something like `ValueError("DataNode 'x': source_format 'directory' requires a directory, got a file at ...")`. A misconfiguration reading as a legitimately empty source is the same silent-success failure mode we flagged in the first review.

factory/outer_loop/population.py
```diff
@@ -196,14 +196,17 @@ def diversity_metric(self) -> float:
...
 
         Returns 0.0 for empty archive, approaches 1.0 as more cells are filled.
+        Uses the first 5 structural axes (depth, fork_degree, agent_count,
+        gate_count, has_data_node) for diversity estimation.
         """
         if not self._grid:
             return 0.0
-        unique_per_axis: list[set[int]] = [set() for _ in range(4)]
+        sample_key = next(iter(self._grid))
+        n_axes = min(len(sample_key), 5)
```
lambdabaa: The docstring says the five axes are "depth, fork_degree, agent_count, gate_count, has_data_node", but `has_data_node` is index 8 of the 9-tuple — index 4 is `_hash_bucket(edge_sig)`, which takes up to 8 values. So this actually uses the edge-signature hash bucket as the fifth diversity axis.

Two consequences: the comment is wrong about what the code does, and the metric's `total_possible` denominator now multiplies by up to 8x, which shifts the scale of a value that feeds diversity-collapse detection (`current < diversity_floor * initial_diversity`). That's a real behavior change hiding inside what reads like a doc fix.

Suggested fix: select the structural axes explicitly rather than taking the first N — e.g. `STRUCTURAL_AXES = (0, 1, 2, 3, 8)` and iterate those indices. That matches the docstring's intent and keeps the metric on structural axes only.

factory/workflow/executor.py
```diff
@@ -628,6 +634,285 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+        if node.split != "all":
+            task_instances = [
+                (item, inst) for item, inst in task_instances
+                if item.metadata.get("split") == node.split
+            ]
+        if node.shuffle:
+            seed = (
+                node.shuffle_seed
+                if node.shuffle_seed is not None
+                else hash(f"{node_id}:{self.run_id}") % (2**32)
```
lambdabaa: Non-blocking: the fallback seed uses `hash()`, which is `PYTHONHASHSEED`-randomized, so it's deterministic within a process but not across restarts — you noted this in the thread, thanks. A `hashlib.sha256(f"{node_id}:{self.run_id}").digest()[:4]` one-liner would make it stable everywhere. Also worth one docstring line: the fallback embeds `run_id`, so only an explicit `shuffle_seed` gives cross-run reproducibility — that's a reasonable default but should be stated.

factory/workflow/executor.py
```diff
@@ -628,6 +634,285 @@ async def throttled_branch(idx: int) -> dict[str, Any]:
...
+                    item_json_path.parent.mkdir(parents=True, exist_ok=True)
+                    item_json_path.write_text(json.dumps(item.model_dump()))
+
+                    try:
+                        item_executor = WorkflowExecutor(
+                            sub_workflow.model_copy(deep=True),
+                            item_project_path,
+                            agent_pool=self.agent_pool,
+                            dry_run=self.dry_run,
+                            initial_context=item.prompt,
```
lambdabaa: Non-blocking: for source_path and inline items, `item.prompt` is `""` by default, so every FnNode-started subgraph logs an `initial_context_ignored` warning once per item. Harmless but noisy at scale (500 items = 500 identical warnings). Passing `initial_context=item.prompt or None` would keep the channel explicit for task_ref items while silencing the noise.

lambdabaa (changes requested):
Re-reviewed after fc3508a8 and the follow-up commits. First: all seven items from the first review are properly fixed, and I verified the important ones empirically rather than just reading the diff — ran a parallelism=2 execution in a scratch git repo (worktrees created, isolated, and cleaned up; each item saw its own `current_item.json`), probed all three source-error tiers (missing path raises, partial JSONL skips bad lines with a warning, all-bad JSONL raises), and confirmed the double-execution edge is rejected by validation. Tests, ruff, and mypy all pass locally. This is a much stronger PR.

Two remaining items block for me, both small, both in the same fails-quietly class as the original findings:

1. **Format/path-kind mismatch is silently treated as empty.** `source_format='directory'` pointed at a file (or `jsonl` at a directory) falls through the `is_dir()`/`is_file()` guards, yields zero items, and the run succeeds with a misleading `data_source_empty` warning. I probed this directly: directory-format-on-file gives `success=True` with zero items. A misconfiguration should raise, not masquerade as a legitimately empty source.

2. **`diversity_metric` doesn't measure what its docstring says.** The docstring claims the five diversity axes are depth, fork_degree, agent_count, gate_count, has_data_node — but `has_data_node` is index 8 of the feature tuple; index 4 is the edge-signature hash bucket (up to 8 values). So the metric's denominator quietly multiplies by up to 8x, and that value feeds diversity-collapse detection (`diversity_floor=0.2`). Select the structural axes explicitly instead of taking the first N.

Three non-blocking notes below. Neither blocking item needs more than a few lines — after those, this is good to merge from my side.

factory/workflow/executor.py
colehurwitz: Fixed in 80a0d349. Added explicit path-kind validation before the is_dir()/is_file() guards: if source_format='directory' but path is a file (or jsonl/csv but path is a directory), raises `ValueError` with a descriptive message like "source_format='directory' but path is a regular file". No more silent fallthrough to zero items.

factory/outer_loop/population.py
colehurwitz: Fixed in 80a0d349. `diversity_metric()` now selects the 5 structural axes explicitly by index — depth(0), fork_degree(1), agent_count(2), gate_count(3), has_data_node(8) — instead of taking `features[:5]` which inadvertently included the edge-signature hash bucket at index 4. Docstring updated to match. The denominator no longer inflates by up to 8x from the hash bucket.

factory/workflow/executor.py
colehurwitz: Fixed in 80a0d349. Fallback seed now uses `hashlib.sha256(f"{node_id}:{self.run_id}".encode()).digest()[:4]` — stable across restarts and PYTHONHASHSEED values. Added docstring noting that the fallback embeds run_id, so only an explicit `shuffle_seed` gives cross-run reproducibility.

factory/workflow/executor.py
colehurwitz: Fixed in 80a0d349. Now passes `initial_context=item.prompt or None` — empty prompts (source_path/inline items) become `None` which skips the initial_context channel entirely, silencing the per-item `initial_context_ignored` warnings. task_ref items with real prompts still flow through as before.

colehurwitz:
## Addressed all items from lambda's second review

Commit `80a0d349` fixes all 5 items. CI is green.

### Blocking (2)

1. **Format/path-kind mismatch** — Added explicit validation before `is_dir()`/`is_file()` guards. `source_format='directory'` on a file (or jsonl/csv on a directory) now raises `ValueError` with a descriptive message. No more silent fallthrough to zero items.

2. **`diversity_metric` axes** — Now selects structural axes explicitly by index (depth=0, fork_degree=1, agent_count=2, gate_count=3, has_data_node=8) instead of `features[:5]`. The edge-signature hash bucket at index 4 no longer inflates the denominator. Docstring updated.

### Non-blocking (3)

3. **Shuffle seed** — Fallback uses `hashlib.sha256` instead of `hash()` — stable across restarts and PYTHONHASHSEED values. Docstring notes that fallback embeds `run_id`; explicit `shuffle_seed` needed for cross-run reproducibility.

4. **`initial_context` noise** — Passes `initial_context=item.prompt or None` — empty prompts become `None`, silencing per-item `initial_context_ignored` warnings for source_path/inline items.

5. **Coverage** — Added tests covering the uncovered `inner_loop.py` lines.

Replied to each inline comment with specifics.

Let me know if this is good to go @lambdabaa :-)

lambdabaa (dismissed):
Took another pass over 80a0d349. Everything I blocked on last round is fixed and I verified the fixes empirically, not just by reading:

**Format/path-kind mismatch** now raises `ValueError` before the `is_dir()` / `is_file()` guards. I ran the executor against both mismatch cases (jsonl format pointing at a directory, directory format pointing at a file) and each fails the workflow with a clear message naming the format and what it got. Legitimately empty jsonl still warns and succeeds, which is the right call since that's a valid dataset state.

**Structural axes**: `STRUCTURAL_AXES = (0, 1, 2, 3, 8)` correctly matches the five structural dimensions of the feature vector, and the hash-bucket + param/prompt axes stay out of the denominator.

**Shuffle seed fallback**: the `sha256(node_id:run_id)` derivation is deterministic per node and per run, exactly what we wanted.

**`initial_context`**: `item.prompt or None` fixes the empty-string prompt.

**`Edge.condition_label`**: nice cleanup, the property reads well and drops the `type: ignore` comments.

Full suite on this commit: 626 passed, ruff and mypy clean.

One residual worth a follow-up (not blocking, it predates this PR): `diversity_metric()` now has a numerator/denominator mismatch. The numerator is `len(self._grid)`, which counts all nine-dimensional cells including hash-bucket distinctions, while the denominator is the structural subspace. Two individuals with identical structure but different edge-hash buckets land in two grid cells and a one-cell structural subspace, so the metric returns 2.0 where the docstring promises it "approaches 1.0". The overflow existed before this change too (the numerator was never structural), and since `_detect_diversity_collapse()` compares current against initial measured the same way, the ratio mostly cancels. The fix is one line if you want it in this PR:

```python
structural_cells = {tuple(key[a] for a in axes) for key in self._grid}
return len(structural_cells) / max(total_possible, 1)
```

Happy to file it as a separate issue otherwise. Thanks for turning this around quickly, the new test coverage is solid.

colehurwitz:
Fixed the `diversity_metric()` numerator/denominator mismatch in `ccfa85b5`.

The numerator now counts unique structural-axis cells instead of all 9-dimensional grid cells:

```python
structural_cells = {tuple(key[a] for a in axes) for key in self._grid}
return len(structural_cells) / max(total_possible, 1)
```

Two individuals with identical structure but different edge-hash buckets no longer inflate the metric above 1.0. Since `_detect_diversity_collapse()` compares current vs initial measured the same way, the ratio behavior is preserved — just the absolute scale is now correct.

Also merged latest main (clean auto-merge, `primitives.py` only).

colehurwitz:
@ceo-review

PR #1487, opened 2026-09-10
workflow: make Workflow.to_dict() deterministic; add OptKnob.description

akashgit:
## Problem

`Workflow.to_dict()` serialized a node's `reads`/`writes` — Python `set`s — in set-iteration order, which is not stable across processes. The same graph could therefore emit different bytes on different runs. For a wire format (`graph.json`) that consumers diff, hash, and compare, that makes the IR unreproducible.

Separately, `OptKnob` had no place to carry a human-readable note about what a knob does.

## Change

- `to_dict()` sorts `reads`/`writes` before serializing, so a given workflow always produces identical bytes.
- `OptKnob.description` (`str = ""`) carries prose for humans; it has no optimizer meaning and is backward-compatible (defaults to empty).

## Verification

- `tests/test_workflow_primitives.py::TestWorkflow::test_to_dict_is_byte_stable` and `test_to_dict_round_trips_via_json` pin the behavior.
- The byte-stability test passes across `PYTHONHASHSEED` 1/2/3/7.
- 68 primitive + package tests pass.

### DIFF
--- a/factory/workflow/package.py
+++ b/factory/workflow/package.py
@@ -58,6 +58,7 @@ class OptKnob(BaseModel):
     the initial ``bounds``.  For prompt knobs this means authoring new
     prompt text; for threshold knobs it means extrapolating the range.
     ``expansion_hint`` tells the optimizer how to generate new values.
+    ``description`` is prose for humans and carries no optimizer meaning.
     """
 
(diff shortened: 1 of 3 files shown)


lambdabaa:
@ceo-review

Issue #1484, opened 2026-09-10
Plugin agent roles can't be used in AgentNode workflow graphs (closed AgentRole enum)


lambdabaa:
## Summary

The plugin system is the official extension story (`docs/plugins.md`), and `PluginRegistry.add_agent_roles()` makes a plugin-defined role valid at the CLI level. But `AgentNode.role` is typed as the closed `AgentRole` enum, so **plugin agent roles cannot be used inside workflow graphs** — which is exactly what the plugin docs promise:

> Plugin workflows can mix plugin-defined agents (`paper-reader`) with built-in agents (`strategist`, `builder`). — `docs/plugins.md:54`

There is currently no supported way to make that sentence true without private-API manipulation.

## Current behavior

1. `AgentRole` (`factory/workflow/primitives.py:16`) is a closed `str` enum with 10 members: `researcher`, `strategist`, `builder`, `health_checker`, `code_reviewer`, `adversarial_tester`, `failure_analyst`, `ceo`, `archivist`, `refiner`.
2. `AgentNode.role: AgentRole` (`factory/workflow/primitives.py:133`) with `ConfigDict(strict=True, extra="forbid")` — constructing an `AgentNode(role="paper-reader")` raises a Pydantic `ValidationError`.
3. `registry.add_agent_roles(["paper-reader"])` (`factory/plugins.py`) only widens CLI validation in `factory/cli/agents.py:169`, which unions `BUILTIN_AGENT_ROLES | plugin_roles`. It does not affect graph validation.

So a plugin role works for `factory agent paper-reader --task ...` but cannot appear in any `AgentNode` of a workflow the plugin registers — including via `Package.compile()`, since composition lowers to the same `AgentNode` IR.

## The workaround plugins are forced into

Plugins currently mutate the enum at import time:

```python
def ensure_plugin_roles() -> None:
    from factory.workflow.primitives import AgentRole

    for name, value in _PLUGIN_ROLES.items():
        if not hasattr(AgentRole, name):
            obj = str.__new__(AgentRole, value)
            obj._name_ = name
            obj._value_ = value
            AgentRole._member_map_[name] = obj
            AgentRole._value2member_map_[value] = obj
            type.__setattr__(AgentRole, name, obj)
```

This works (it's what we run in production at refactory-midstream, which registers 7 domain-specific roles this way), but:

- It manipulates `_member_map_` / `_value2member_map_` — private CPython `Enum` internals that can change without notice.
[...]

PR #1491, opened 2026-09-10
feat: first-class plugin workflow registration

### LINKED ISSUE #1490: First-class plugin workflow registration: wire PluginRegistry into WorkflowRegistry, define discovery priority, fix skill-cache invalidation
## Summary

The plugin system is the official extension story (`docs/plugins.md`), and `PluginRegistry.add_workflow_search_path()` exists so plugins can contribute workflow graphs. But the two halves are not wired together, and the workflow discovery layer has no defined semantics for plugins. The result: **any plugin that contributes workflows today has to monkey-patch the engine**, and we carry those patches in production.

This issue covers two of the three patches a real plugin (our midstream plugin, which registers 20+ workflows) currently needs. The third, the `AgentRole` enum, is filed separately as #1484.

## The patches a plugin needs today

**Patch 1: manual bridge.** `PluginRegistry.add_workflow_search_path()` records the path, but `WorkflowRegistry.discover()` never consults `PluginRegistry`. The only way a plugin's paths take effect is appending them to the private class attribute `WorkflowRegistry._search_paths` by hand. Every plugin has to know this internal.

**Patch 2: rewritten `discover()`.** Our plugin replaces `WorkflowRegistry.discover` with its own classmethod because upstream's layering is wrong for plugins. Two bugs, both verifiable in `factory/workflow/registry.py`:

### Bug A: the layering is backwards once plugins exist

`discover()` runs layers in this order (`registry.py:87-104`):

1. builtins (`_load_builtins`)
2. user-global (`~/.factory/workflows/`)
3. project-local (`.factory/workflows/`)
4. non-user paths on `_search_paths` (the only channel a plugin has)

Each layer calls `_discover_in_directory`, which assigns `cls._entries[name] = ...` **unconditionally** (`registry.py:150`), so the last layer to claim a name wins. Consequence: a plugin-registered path (layer 4) **overrides project-local workflows** (layer 3). That is inverted. Project-local is the user's explicit override for the project in front of them; a plugin should never shadow it.

The intended priority is presumably: project-local > plugin > user-global > builtin. Nothing in the code enforces or tests that today.

### Bug B: shadowing a builtin is silent

The shadow warning in `_discover_in_directory` only fires when the previous entry is *not* a builtin (`registry.py:143-149`). So a plugin or project workflow that overrides `improve` or `build` does so without a word. With plugins now the official extension surface, a silent collision with a built-in mode is a footgun: behavior changes depending on plugin install order and nothing tells anyone why.

**Patch 3: skill-cache checksum.** Our plugin also patches `skill_cache._compute_checksum` (`factory/skill_cache.py:30`) because editing a workflow `.py` file did not invalidate the cached exported skills. Root cause: the checksum hashes only `model_dump()` of the constructed `Workflow` objects, but the exported `SKILL.md` also depends on entry-level data such as the `meta` description, which is not a field on the `Workflow` model. The plugin's fix folds SHA-256 of each discovered workflow's source file into the checksum, so any content change invalidates the cache.

## Proposal

Not a new registry. The existing pieces are close to right; what is missing is connecting them and defining semantics:

1. **`WorkflowRegistry.discover()` consumes `PluginRegistry.workflow_search_paths` natively.** A plugin calls `add_workflow_search_path()` in its `register()` function and is done. No private-attribute bridge.
2. **Define, document, and test the priority order**: project-local > plugin > user-global > builtin. Implementation should guard overwrites so a lower-priority source cannot clobber a higher one, and log a warning on every shadow event, including over builtins. (Decide whether a plugin shadowing a builtin should warn or be rejected outright; project-local shadowing a builtin is legitimate and should stay allowed, with a warning.)
3. **Fix the skill-cache checksum** so it covers everything the export reads. Either fold entry-level metadata into the hashed payload, or take the plugin's approach and hash the source `.py` file contents of every discovered workflow. The latter is strictly conservative: any edit to a workflow file invalidates the cache.

## What this does not cover

- `PluginRegistry.add_modes()` works correctly today without patching. The pain is entirely in mode-to-workflow resolution (discovery) and skill export (cache invalidation), so this issue is scoped to `factory/workflow/registry.py` and `factory/skill_cache.py`.
- The `AgentRole` enum extension is #1484.

## Acceptance criteria

A plugin that calls `registry.add_workflow_search_path(...)` in `register()`:

- has its workflows discovered with no manual `_search_paths` bridge,
- does not shadow project-local workflows,
- sees a warning when it shadows a builtin,
- and its users get regenerated SKILL.md exports when its workflow files change.

Stated as a migration test: our plugin (midstream) can delete its `_registry_patch.py` and its `_search_paths` bridge, and its resolution order (project > plugin > user > builtin) and cache-invalidation behavior are preserved by upstream tests.

lambdabaa:
## What this does

Closes #1490. Three pieces, one theme: plugins that register workflows should work without downstream repos monkey-patching the registry.

**1. Native plugin path consumption.** `WorkflowRegistry.discover()` now reads `PluginRegistry.workflow_search_paths` directly. If midstream's lightwell plugin calls `factory.plugins.get_registry().add_workflow_search_path(...)`, discovery picks up its workflows with no manual `_search_paths` bridge and no `discover` monkey-patch. If the plugin registry is unavailable for any reason, discovery logs at debug and continues without plugin workflows.

**2. Discovery priority.** Sources are now layered explicitly: **project > plugin > user > builtin**. Previously, `_entries[name]` was overwritten unconditionally, so later layers always won. Because discovery order puts plugin paths after user paths but before project-local discovery, plugin workflows appended to `_search_paths` silently overrode project-local ones with the same name. That inversion is fixed: a project-local `improve.py` now beats a plugin workflow named `improve`. Every shadow event is logged (warning when the new source wins, info when it's skipped), including builtin shadowing, which previously went unlogged.

**3. Skill-cache invalidation on source edits.** `_compute_checksum` now hashes the discovered workflow source files alongside the model dump. Before, a comment-only edit to a plugin workflow file left the checksum unchanged, so the exported SKILL.md cache went stale. The `source_files` parameter is optional with the old behavior as default, so existing callers are unaffected.

## Notes for reviewers

- The priority map is a module-level `_SOURCE_PRIORITY` dict with a small `_source_priority()` helper that defaults unknown labels to the plugin rank, so a new source label degrades gracefully rather than erroring.
- The two failures you may see in `test_compose.py` (`TestTomlTaskCapabilities::test_chess_evolve_toml_no_builder_required` and `test_chess_evolve_toml_passes_any_workflow`) are pre-existing on `main` when those files run after `test_a*`/`test_b*`. I verified by stashing my changes and re-running the same selection on base. They're an ordering interaction unrelated to this PR.

## Testing

18 new tests in `tests/test_workflow_registry_plugins.py` covering plugin path consumption, all priority combinations (project shadows plugin, plugin shadows user, user shadows builtin, legacy `_search_paths` back-compat, legacy-does-not-shadow-project regression), shadow-warning logging, checksum source-file sensitivity, and an end-to-end skill-cache miss/hit/miss cycle on a comment-only file edit. All 75 tests across the affected suites pass, plus ruff and mypy clean.

Downstream context: this PR plus #1484 together retire all four monkey-patches midstream's lightwell plugin currently applies (see #1490 for the inventory).

### DIFF
--- a/factory/skill_cache.py
+++ b/factory/skill_cache.py
@@ -27,16 +27,33 @@ def _sort_recursive(obj: object) -> Any:
     return obj
 
 
-def _compute_checksum(workflows: dict[str, Workflow]) -> str:
-    """Deterministic checksum from workflow Pydantic models.
+def _compute_checksum(
+    workflows: dict[str, Workflow],
+    source_files: dict[str, str] | None = None,
+) -> str:
+    """Deterministic checksum from workflow Pydantic models + source files.
 
     Serialises all workflows via model_dump(mode='json'), sorts by name,
     then SHA-256 hashes the canonical JSON.  Returns the first 16 hex chars.
+
+    When *source_files* maps workflow name → path of the .py file it was
+    discovered from, the SHA-256 of each file's bytes is folded into the
+    checksum.  This makes the cache key cover everything the skill export
+    reads: the exported SKILL.md depends on entry-level data (e.g. the
+    ``meta`` description) that is not a field on the Workflow model, so
+    hashing models alone can serve stale skills after a file edit.
     """
     payload = {name: wf.model_dump(mode="json") for name, wf in sorted(workflows.items())}
     payload = _sort_recursive(payload)
     blob = json.dumps(payload, sort_keys=True).encode()
-    return hashlib.sha256(blob).hexdigest()[:16]
+    h = hashlib.sha256(blob)
+    for name, path_str in sorted((source_files or {}).items()):
+        try:
+            h.update(name.encode())
+            h.update(Path(path_str).read_bytes())
+        except OSError as exc:
+            log.debug("skill_cache.checksum_read_failed", path=path_str, error=str(exc))
+    return h.hexdigest()[:16]
 
 
 def ensure_skills(project_dir: Path, *, mode: str | None = None) -> list[Path]:
@@ -64,6 +81,7 @@ def _ensure_skills_inner(project_dir: Path, *, mode: str | None = None) -> list[
 
     builtin_workflows: dict[str, Workflow] = {}
     project_workflows: dict[str, Workflow] = {}
+    source_files: dict[str, str] = {}
 
     for name, entry in entries.items():
         wf = WorkflowRegistry.get_workflow(name, project_dir)
@@ -73,10 +91,12 @@ def _ensure_skills_inner(project_dir: Path, *, mode: str | None = None) -> list[
             project_workflows[name] = wf
         else:
             builtin_workflows[name] = wf
+            if entry.path != "<builtin>":
+                source_files[name] = entry.path
 
     log.info("skill_cache.project_workflows_discovered", count=len(project_workflows))
 
-    checksum = _compute_checksum(builtin_workflows)
+    checksum = _compute_checksum(builtin_workflows, source_files)
 
     cache_dir = Path.home() / ".factory" / "cache" / "skills" / checksum
     skills_target = project_dir / "skills"
--- a/factory/workflow/registry.py
+++ b/factory/workflow/registry.py
@@ -130,8 +130,44 @@ class WorkflowRegistry:
             if project_wf_dir.is_dir():
                 cls._discover_in_directory(str(project_wf_dir), "project")
 
+        cls._warn_mode_drift()
+
         log.info("workflow_registry.discovered", count=len(cls._entries))
         return cls._entries
+
+    @classmethod
+    def _warn_mode_drift(cls) -> None:
+        """Log consistency warnings between plugin modes and discovered workflows.
+
+        A plugin mode with no discovered workflow of the same name usually means
+        a typo or a renamed file — without a warning it surfaces only as a
+        silent fallback to the default improve loop. A plugin workflow that is
+        not declared as a mode is legal (subgraph libraries, composed
+        packages), so that direction is info-level only.
+        """
+        try:
+            from factory.plugins import get_registry
+
+            plugin_modes = set(get_registry().modes)
+        except Exception as exc:
+            log.debug("workflow_registry.plugin_modes_unavailable", error=str(exc))
+            return
+        if not plugin_modes:
+            return
+
+        for mode in sorted(plugin_modes):
+            if mode not in cls._entries:
+                log.warning(
+                    "workflow_registry.mode_without_workflow",
+                    mode=mode,
+                    action="ceo_falls_back_to_improve_loop",
+                )
+        for name, entry in cls._entries.items():
+            if entry.source == "plugin" and name not in plugin_modes:
+                log.info(
+                    "workflow_registry.plugin_workflow_not_a_mode",
+                    name=name,
+                )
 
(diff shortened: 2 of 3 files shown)


lambdabaa:
@ceo-review

shivchander:
This is a clean fix — native plugin-path consumption, explicit priority, and source-file checksums cover the discovery pain well. A few adjacent gaps that this doesn't close (may be intentionally out of scope, flagging for completeness):

**1. Workflow discovery and mode/CLI exposure are still two separate registrations.**
Say a plugin registers a workflow search path and drops `foo.py` in it. With this PR, `foo` is discovered and resolvable — but to expose it as an invokable mode (`--mode foo`) and have it appear in the mode listing, the plugin still has to register the name through a *second* call (`add_modes`). The two can drift: a workflow that's discoverable but not invokable, or a declared mode name with no backing workflow. Could discovery optionally surface discovered workflows as modes, or could the two share a source of truth?

**2. No discovery-time consistency check for "declared but not discoverable" (and vice-versa).**
If a plugin declares mode `bar` but no `bar` workflow is discoverable (typo, renamed file), it surfaces only at invocation as a confusing "unknown mode." A load/discovery-time warning for *declared-but-missing* and *discovered-but-undeclared* would catch this early — a natural complement to the shadow logging this PR already adds.

**3. Worth confirming the user-facing `list` path.** `discover()` now populates entries natively and skill-cache/`show` go through it — is the `workflow list` command reading the same registry, so plugin workflows show up there too and not only via `show`?

lambdabaa:
Thanks Shiv — all three points landed or answered in da02ebe9.

**2. Discovery-time consistency warnings — done.** `discover()` now ends with a drift check between plugin-registered modes and discovered workflows:
- Declared mode with no discovered workflow of the same name → `workflow_registry.mode_without_workflow` warning. This is the typo/renamed-file case you described; without it, the failure was a silent fallback to the improve loop.
- Discovered plugin workflow not declared as a mode → `plugin_workflow_not_a_mode` at info level, since that's legal (subgraph libraries, composed packages).
- No plugin modes declared → check skipped entirely; nothing to drift against.

One deliberate exclusion: outer-loop ephemeral modes (`evolve-gen{N}-...`) register through `.factory/workflows/`, not `PluginRegistry.modes`, so they can't false-positive this check.

**1. Two separate registrations — agreed, and left alone for now.** I don't think auto-surfacing discovered workflows as modes is right: a plugin shipping a library of subgraph workflows would suddenly expose every one as a `--mode`, and mode-without-workflow is a legal configuration by design (the CEO improve-loop fallback). The drift warnings above close the dangerous half of the drift (declared but broken); the harmless half (workflow that could be a mode but isn't) stays info-level. If you want the single-source-of-truth version, that feels like its own issue — happy to file one.

**3. `workflow list` — confirmed, same registry.** `_cmd_list` goes through `WorkflowRegistry.list_workflows(project_path)` → `discover()`, which now consumes plugin paths natively, so plugin workflows show up in `factory workflow list` exactly as they do in `show` and the skill cache.

lambdabaa:
@ceo-review

PR #1492, opened 2026-09-10
feat: sanctioned AgentRole extension for plugin agent roles

### LINKED ISSUE #1484: Plugin agent roles can't be used in AgentNode workflow graphs (closed AgentRole enum)

lambdabaa:
Closes #1484.

## What this does

Gives plugins a sanctioned way to extend the `AgentRole` enum, so plugin-defined roles finally work inside workflow graphs — the thing `docs/plugins.md` has been promising all along ("plugin workflows can mix plugin-defined agents with built-in agents").

**The public API.** `factory.workflow.primitives.register_agent_role("paper-reader")` adds `AgentRole.PAPER_READER` (member name derived from the value, or given explicitly). The role is then valid in `AgentNode.role`, `GateNode.evaluator_role`, and `AgentConfig.role`. Registration is idempotent, and collisions raise `ValueError` with a clear message: builtin value, builtin member name, value already taken, name already taken.

**Registry wiring.** `PluginRegistry.add_agent_roles(["paper-reader"])` now calls `register_agent_role()`, so a plugin that registers roles for the CLI gets graph-usable roles for free — no enum mutation in the plugin's `register()`. The registry keeps its existing skip-with-warning behavior on builtin collisions (builtins always win), consistent with `add_commands` and `add_modes`.

## The latent bug this fixes

Downstream plugins (we run one in production) work around the closed enum by inserting members into `AgentRole._member_map_` / `_value2member_map_` at import time. That makes in-process construction work, but it hides a real bug: **Pydantic freezes an enum's valid values into model core schemas at class-definition time.** A dynamically added member serializes to JSON fine, but `model_validate_json` of a workflow containing it fails with `Input should be 'researcher', 'strategist', ...` — the frozen schema never heard of the new role. So ephemeral mode JSONs, outer-loop candidate serialization, anything that roundtrips workflow IR through JSON, silently breaks for plugin roles.

I verified this empirically on `main`: mutate the enum the way the workaround does, `Workflow.model_validate_json(wf.model_dump_json())` raises 59 validation errors. After `register_agent_role()` rebuilds the schemas (it walks every `BaseModel` defined in `primitives.py`, children before the containers that embed them), the same roundtrip is clean.

## Notes for reviewers

- The schema rebuild iterates `vars()` of `primitives.py` in definition order, which visits node classes before `Workflow`/`Factory` — the order the nested schema graph needs.
- `_registered_plugin_roles` (keyed by value) is what makes re-registration idempotent while a builtin collision still raises; without it, `register_agent_role("builder")` would happily return `AgentRole.BUILDER` and look like success.
- Strict-mode string construction (`AgentNode(role="paper-reader")`) fails exactly as it does for builtins (`AgentNode(role="builder")` also raises in strict mode) — plugin roles now have parity with builtins everywhere builtins work: enum members in strict mode, strings in lax mode (`from_dict`), and strings in JSON.

## Testing

24 tests in `tests/test_plugin_agent_roles.py`: registration semantics (derived/custom names, idempotency, all four collision cases, identifier validation), graph usage (AgentNode/GateNode/AgentConfig, JSON and to_dict/from_dict roundtrips, builtin roles still validating after a registration), skill export rendering (`factory agent paper-reader` appears in the generated SKILL.md), and registry integration (graph-usable roles via `add_agent_roles`, builtin-skip warning, duplicates, invalid roles). An autouse fixture snapshots and restores the enum between tests, since registration is process-global — without it roles would leak into the outer loop's `random.choice(list(AgentRole))` and friends. Full related suites (819 tests across workflow, skill, plugin, and outer_loop) pass, plus ruff and mypy clean.

For midstream specifically: `ensure_cve_roles()` in `lightwell/workflows/_helpers.py` can be replaced by seven `register_agent_role()` calls (or just the existing `add_agent_roles()` list, if it's already registered as a plugin), and the JSON roundtrip bug it was carrying goes away with it.


lambdabaa:
@ceo-review

shivchander:
The schema-rebuild fix is the important part here — the frozen-enum JSON-roundtrip failure is a nasty silent one, glad it's tackled head-on. One adjacent gap worth noting (possibly out of scope):

**Role registration makes a role *valid*, but not *behaved*.**
A plugin calls `register_agent_role("paper-reader")` and uses it on a node — the graph builds and JSON roundtrips cleanly. But at run time that role still needs its prompt/identity resolved from somewhere, and shipping/resolving that is left entirely to the plugin — and is easy to get wrong across install methods (e.g. the prompt data files not making it into the built wheel, so the role silently resolves to "no instructions"). It would round out the "plugin-defined agents are first-class" story to have a sanctioned way to register a role's prompt/identity *alongside* the role (parallel to how this PR sanctions the enum), or at least a documented resolution contract plus a warning when a registered role has no resolvable prompt.

lambdabaa:
Thanks Shiv — took the "at least" version in fb36b95f: the documented resolution contract plus warnings on both ends. The full "register a prompt alongside the role" API I've left as a possible follow-up, reasons below.

**What landed:**

- **Documented contract.** `docs/plugins.md` now spells out the three-tier resolution order for plugin roles: project override (`.factory/agents/<role>.md`) → user-global (`~/.factory/agents/prompts/<role>.md`) → factory default (builtins only). A plugin role must ship a prompt and install it to tier 2 on first load, or rely on projects providing tier 1.
- **Load-time warning.** After all plugins register, `_warn_missing_role_prompts()` checks every registered role against tiers 2 and 3 and logs `plugin_agent_role_prompt_missing` when neither exists — this is exactly your wheel-packaging case, caught before any CEO cycle starts rather than partway through one. The warning notes that a project override also satisfies the role, so it's not a false alarm for setups that rely on tier 1.
- **Better runtime error.** One correction to the premise: a role with no resolvable prompt doesn't silently resolve to "no instructions" today — `resolve_prompt` raises `FileNotFoundError`. But the old error didn't say anything useful about *why* a role could be valid yet promptless. It now appends a hint when the role is plugin-registered, pointing at the plugin's data files as the likely cause.

**Why not the full prompt-registration API yet:** the interesting design question is where a registered prompt would live in the resolution order. Writing to `~/.factory` at registration time is an install side effect from library code, and slotting plugin prompts into the three-tier order (below or above project overrides?) changes behavior for existing setups. That deserves its own discussion rather than a rider on this PR — happy to file an issue if you want to pursue it.

lambdabaa:
@ceo-review

RohanAwhad:
## Code Review

**Decision: `ready_with_followups`**

> **Scope**: Reviewed 2 unique commits out of 2 total (no upstream sync). Target branch: `main`. Diff: +479 −2 across 6 files. Non-test source ≈166 lines.

### Intent & Optimality (pr-intent-review)
**Understood intent**: Implement issue #1484's **Option B (first-class extension API)** — a sanctioned `register_agent_role()` in `factory/workflow/primitives.py` that extends the closed `AgentRole(str, Enum)` and rebuilds Pydantic core schemas so plugin roles validate in `AgentNode.role`/`GateNode.evaluator_role`/`AgentConfig.role` and roundtrip through JSON and `to_dict`/`from_dict`. `PluginRegistry.add_agent_roles()` now routes through it, and two prompt-resolution safety nets (load-time warning + invocation-time hint) were added in the review-driven second commit.

**Optimality**: optimal for the chosen direction. The issue explicitly weighed Option A (registry-driven validation, makes deserialization depend on plugin load state), Option B (sanctioned enum extension), and Option C (string roles everywhere). Option B keeps the enum closed to Pydantic (strict typing preserved), centralizes the private-API insertion plugins were already doing in production, adds collision protection + idempotency the workaround lacked, and — critically — adds the schema rebuild that turns "in-process construction works" into "JSON roundtrips work." All four acceptance criteria in #1484 are met (validate, serialize roundtrip, builtin collision rejected, skill export renders).

**Consistency with codebase**: `add_agent_roles` collision handling mirrors `add_commands`/`add_modes` (skip-builtin-with-warning, keep-first-on-duplicate) — `factory/plugins.py:81`. `BUILTIN_AGENT_ROLES` (`_parser_groups.py:6`) exactly matches the enum values, no drift. The enum-mutation moves are identical to the issue's documented workaround, now centralized and hardened.

### Breaking Changes (code-review-breaking-change)
See Action Items 1 and 2. No always-fatal API/CLI/config break; the two risks both stem from `AgentRole` no longer being closed at process scope.

### Change Size (code-review-change-size)
PASS. Non-test source ≈166 lines — well under the 500/800 thresholds. Healthy ~2:1 test-to-code ratio. **Not scope creep**: (A) `register_agent_role` is dead code without (B) the registry wiring, and (C) the prompt safety nets were added in direct response to reviewer @shivchander's feedback and are framed as the other half of one contract ("valid" vs. "behaved"). Cohesive; no split warranted.

### Test Integrity (code-review-test-integrity)
No integrity issues. Verified against merge-base `e8c12ab`: one new test file added (`tests/test_plugin_agent_roles.py`), **zero** deletions or modifications to existing tests, **no** CI/workflow files touched, no skips/xfails, no coverage weakening. See Action Item 3 for a minor test-quality note and Action Item 5 for a doc-count discrepancy.

### Comment Density (code-review-comment-density)
Effectively clean. The added commentary is legitimate "why" documentation (Pydantic frozen-schema behavior, `vars()` iteration order, `_registered_plugin_roles` rationale, three-tier resolution). One nit: `_is_plugin_role` docstring at `factory/agents/runner.py:43` largely restates the signature — borderline, fine to keep.

### Race Condition (code-review-race-condition-check)
No real race. `register_agent_role` and `_rebuild_role_schemas` mutate process-global state without a lock, but they are reachable only via `load_plugins()`, which runs serially on the main thread inside `build_parser()` before any command handler and before the codebase's real concurrency (ThreadPoolExecutor in `evaluator.py`, `asyncio.gather` in `executor.py`), all of which only **read** `AgentRole`. See Action Item 4 for a forward-looking guard suggestion. No filesystem/temp-path races introduced.

### Spec Compliance
- **PRD**: none found (`.dingllm/` absent). Source of truth is **issue #1484** — implementation matches its Option B and satisfies all listed acceptance criteria.
- **Docs**: `docs/plugins.md` — updated and aligned; documents the three-tier resolution contract and both safety nets.

---

### Action Items

- [ ] **1. [MEDIUM] Cross-process deserialization of plugin-role workflows depends on plugin load state** — `factory/workflow/primitives.py:369`
<details>
<summary>Details</summary>

**Spec gap** (the issue flagged this exact tradeoff for Option A, but it applies to Option B too and is not documented for the shipped design).

`Workflow.from_dict` → `node_cls.model_validate(node_data, strict=False)` (`primitives.py:369`) is called to rehydrate persisted workflows in `factory/outer_loop/filesystem.py:193`, `factory/outer_loop/engine.py:260/275/296/377/497`, and `factory/outer_loop/mode_registry.py:70/141`. A workflow persisted (as dict/JSON) **with** a plugin loaded will fail to deserialize in any process where that plugin — and thus `register_agent_role()` — has not run, because the plugin role string is not a valid `AgentRole` member there. There is no graceful unknown-role fallback.

This is inherent to Option B and is **not a regression** (plugin roles could not appear in graphs at all before this PR), so it is not a merge blocker. But it is a real operational footgun for the outer loop / ephemeral-mode persistence. Recommend: (a) document in `docs/plugins.md` that plugin-role workflows are only deserializable in processes where the plugin is loaded, and/or (b) consider a clear error message when `from_dict` encounters an unknown role that names plugin-load as the likely cause (parallel to the `resolve_prompt` hint this PR already added).
</details>

- [ ] **2. [LOW] Plugin roles leak into unrelated outer-loop workflows via `random.choice(list(AgentRole))`** — `factory/outer_loop/mutations.py:1026,1054`, `factory/outer_loop/designer.py:285`
<details>
<summary>Details</summary>

**Outside spec scope** (design consequence, not a spec violation).

Because `register_agent_role` appends to `_member_names_`/`_member_map_`/`_value2member_map_`, `list(AgentRole)`/`set(AgentRole)` now include plugin roles process-wide. If plugin registration and the outer loop ever run in the same process, `mutations.py:1026`/`1054` can inject a plugin role into an unrelated workflow, producing a node whose prompt does not resolve (`FileNotFoundError` at invocation — not caught by the new load-time warning, which only scans registered roles that lack prompts). The PR's own test fixture (`_restore_agent_role`, test lines 21–49) acknowledges this leak within the test suite. In production these typically run in separate processes, bounding the impact — hence LOW. Worth a note or a guard if same-process operation is possible.
</details>

- [ ] **3. [LOW] Prefer `pytest.raises(match=...)` over `try/except/else + pytest.fail`** — `tests/test_plugin_agent_roles.py:281`
<details>
<summary>Details</summary>

**Outside spec scope** (test quality).

`test_resolve_prompt_error_mentions_plugin_registration` uses a `try/except FileNotFoundError ... else: pytest.fail(...)` pattern. The sibling test at lines 309–311 already uses the cleaner `with pytest.raises(FileNotFoundError) as excinfo:` form. Aligning the two improves robustness and readability. Also: the teardown in `_restore_agent_role` (lines 37–49) has no `try/finally`, so an exception mid-restore corrupts enum state for later tests; and the `isupper()` heuristic at line 33 would miss custom lowercase/mixedCase member names that the API otherwise permits. Minor.
</details>

- [ ] **4. [INFO] Add a single-threaded-startup invariant note/guard to `register_agent_role`** — `factory/workflow/primitives.py:197`
<details>
<summary>Details</summary>

**Outside spec scope** (defensive hardening).

The unlocked global mutation is safe **only** because plugin loading is serial on the main thread at startup. A brief comment documenting this invariant (or an assertion/lock) at `register_agent_role` would protect it against a future change that loads plugins off-thread. No action needed for correctness today.
</details>

- [ ] **5. [INFO] PR body says "24 tests"; the file contains 29** — `tests/test_plugin_agent_roles.py`
<details>
<summary>Details</summary>

**Outside spec scope** (doc accuracy). `grep -c 'def test_'` returns 29. Trivial; update the PR description if desired.
</details>

---

> **Verification note**: This review environment lacks `uv`/`pip`/`pytest` and cannot import `pydantic`/`structlog`, so the 29 tests and the "59 validation errors on `main`" claim could not be executed here. All findings are static-verified against the actual source (callers cross-referenced, negative-evidence greps performed). Reviewers with a working env should confirm via `uv sync && uv run pytest tests/test_plugin_agent_roles.py -v` plus `ruff check .` and `mypy factory/`.

---
— Reviewed by Coding Agent

PR #1507, opened 2026-09-15
DesignerAgent: inject DataNode subgraph nodes alongside frozen DataNode

### LINKED ISSUE #1503: DesignerAgent: inject DataNode subgraph nodes alongside frozen DataNode
## Problem

When `_inject_frozen_nodes()` in `factory/outer_loop/designer.py` copies a frozen DataNode into a designer variant, it copies only the DataNode itself — not the subgraph nodes it references (`subgraph_entry`, `subgraph_exit`, and all nodes between them).

The result: the variant has a DataNode pointing to `subgraph_entry="generator"` but `generator` doesn't exist in the variant's nodes dict. The variant fails `frozen_node_violated` validation and scores 0 every time.

This wastes 2 population slots and ~20 min of evaluation budget per generation on candidates guaranteed to fail.

## Root cause

`_inject_frozen_nodes()` treats each frozen node as independent. But a DataNode is a **composite** — it needs its subgraph (entry node → intermediate nodes → exit node + all edges between them) to function. Injecting just the DataNode without its subgraph is like copying a function declaration without its body.

## Fix

When `_inject_frozen_nodes()` encounters a frozen `DataNode`, it should also:
1. Use `_collect_subgraph_nodes(seed_workflow, node.subgraph_entry, node.subgraph_exit)` to find all subgraph nodes
2. Copy those nodes from the seed workflow into the variant's nodes dict
3. Copy the edges between those nodes from the seed workflow

This is the same pattern as `_auto_frozen_nodes()` (which already expands frozen sets to include subgraph nodes for mutation protection) — just applied to the designer's injection path.

## Evidence

From chess-evolve outer loop run:
```
generation 0: frozen_node_violated workflow=minimal_chessevolve → score 0
generation 1: frozen_node_violated workflow=minimal_chessevolve → score 0
generation 2: frozen_node_violated workflow=minimal_chessevolve → score 0
generation 3: frozen_node_violated workflow=minimal_chessevolve → score 0
```

Every generation, 1-2 designer variants waste budget scoring 0.

colehurwitz:
Closes #1503

## Changes

- **`_inject_frozen_nodes()` extended** to detect frozen DataNodes and inject their complete subgraph (nodes + edges) from the seed workflow using `_collect_subgraph_nodes()`
  - Node dedup: skips subgraph nodes already present in the variant
  - Edge dedup: uses `(source, target, condition)` signature set to avoid duplicates
  - Validates subgraph_entry/exit exist in seed before expansion; logs warning and skips if missing
  - Documents nested DataNode non-recursion as a known limitation
- **All 3 call sites updated** (`design_minimal`, `design_thorough`, `design_custom`): edges list defined before `_inject_frozen_nodes()` call and passed as second parameter
- **`_rewire_data_nodes()` fix**: excludes injected subgraph nodes from terminal candidate selection to prevent non-deterministic rewiring failures
- **5 new tests** in `TestInjectFrozenDataNodeSubgraph`:
  1. Multi-node subgraph injection (3 nodes, 2 edges)
  2. No duplicate nodes when frozen_node_ids overlaps subgraph
  3. No duplicate edges when template and seed share an edge
  4. Missing subgraph_entry logs warning, no crash
  5. Post-injection variant passes structural validation
- **2 pre-existing tests updated** to filter expected unreachable-node warnings from dead subgraph nodes


colehurwitz:
@ceo-review

PR #1509, opened 2026-09-15
feat: task-setup mode for scaffolding new Task classes

### LINKED ISSUE #1506: feat: task-setup mode for scaffolding new Task classes
## Problem

Creating a new Task (the 4-hook interface: `instances()`, `setup()`, `prompt()`, `verify()`) requires the user to write Python from scratch with no guidance. There's no scaffold command, no interactive setup, no way to generate a starter Task from a description of the domain.

## Proposal: `--mode task-setup`

A standalone mode that scaffolds a new Task class through interactive research and generation.

```bash
factory ceo --mode task-setup --focus "chess engine evaluation against Stockfish"
factory ceo --mode task-setup --focus "optimize sorting algorithm for speed"
factory ceo --mode task-setup --focus "drug candidate binding affinity scoring"
```

### What it does

1. **Research** — CEO asks about the domain: what to measure, how to verify, what instances to evaluate against
2. **Generate** — produces either:
   - **TOML** for simple tasks (verify = shell command, scoring = exit code or regex)
   - **Python scaffold** for complex tasks (custom evaluation logic, multi-stage verification)
3. **Validate** — runs the generated Task to confirm `instances()` returns items and `verify()` is callable
4. **Output** — a ready-to-use Task file that works with `--task` in create mode and `task_ref` in DataNode

### Workflow graph

`fork(researcher_domain, researcher_verification) → join → CEO gate → strategist → user gate → builder → FnNode(validate_task) → archivist`

10 nodes, 12 edges. The parallel researchers investigate the domain and verification methods independently before the strategist synthesizes a plan.

### Relationship to other issues

- **#1505** (`--data-node` and `--task` flags) — task-setup produces the Task that `--task` consumes. These are complementary: task-setup creates the Task, create mode creates the workflow that uses it.
- **PR #1481** — original implementation of task-setup mode. The mode design is solid; it should be extracted and landed independently of the create mode changes.

### Not in scope

- OptKnob auto-generation (belongs in #1505's `--task` directive injection)
- DataNode creation (belongs in #1505's `--data-node` flag)
- Task discovery/resolution (`resolve_task()` — already implemented in #1481, land separately)

Credit: @akashgit for the original task-setup mode design in #1478/#1481.

colehurwitz:
Closes #1506

## Changes

- **Created `.factory/workflows/task-setup.py`** — portable workflow file with v2 director topology (9 nodes, 11 edges) for scaffolding new Task classes through interactive research, strategy synthesis, and generation
- **Research Director** (CEO) dynamically spawns 3+ researchers: domain, verification, and outer loop feedback (mandatory)
- **Strategy Director** (CEO) spawns 2+ strategists: architecture and verification/evolution (mandatory)
- **QA Director** (CEO) spawns 3 adversarial testers: structural validation, evolution-readiness (mandatory), domain correctness
- **User gate** after strategy for plan approval; CEO gates after research and QA with max_iterations=2
- **FnNode** runs `factory task validate` with blocking=True
- **Non-blocking archivist** for session archival
- **Prompt adjustment** per user feedback: uses generic "reflection and improvement mechanisms" framing instead of hardcoded reflector algorithm references ("contrastive analysis", "top-K vs bottom-K"). Keeps design guidance (≥4 numeric keys, categorical error classification, per-instance breakdowns, chess-evolve example)
- **38 tests** in `tests/test_workflow_task_setup.py` covering graph structure, node types, exact topology (11 edges), prompt quality, artifact checks, and registry discovery

## Verification

- `factory workflow validate --file .factory/workflows/task-setup.py` → VALID (9 nodes, 11 edges)
- `factory workflow export-skills` → workflow-task-setup/SKILL.md generated
- Comprehensive verification script from spec → ALL CRITERIA PASSED
- All 38 tests pass, ruff check clean


.factory/workflows/task-setup.py
```diff
@@ -0,0 +1,1110 @@
...
+
+meta = {
+    "name": "<task_name>",
+    "description": "<description>",
+}
+
+
+def task() -> <TaskClassName>:
+    return <TaskClassName>()
+```
```
georgosgeorgos, line 658: **[Critical] validate_task failure permanently halts the workflow — no recovery path**

When `factory task validate` exits non-zero, `_execute_action_node` catches the `RuntimeError` and sets `halted=True`, then returns without following the unconditional edge to `archivist`. There is no RELOOP edge back to `builder`. Any task file that passes agent QA but fails the CLI validator forces a full 9-node workflow restart.

**Fix:** Convert `validate_task` to a `GateNode(evaluator_type="fn")` with the `factory task validate` command, plus a RELOOP edge back to `builder` so failed validation sends the builder back to fix the issues before re-validating.

```python
"validate_task": GateNode(
    id="validate_task",
    evaluator_type="fn",
    evaluator_command=_VALIDATE_TASK_COMMAND,
    gate_prompt="Task validation failed. Return to builder to fix the issues.",
    reads={".factory/generated-task-name.txt"},
),
# and add:
Edge(source="validate_task", target="builder", condition=VerdictType.RELOOP),
Edge(source="validate_task", target="archivist", condition=VerdictType.PROCEED),
```

.factory/workflows/task-setup.py
```diff
@@ -0,0 +1,1110 @@
...
+
+## Critical Rules
+
+1. **TOML vs Python**: Choose TOML only when verify is a shell command AND
+   instances come from a directory AND prompts are static. Otherwise Python.
+   When in doubt, choose Python — it's always capable enough.
+2. **Name consistency**: task_name is kebab-case. meta['name'] and
+   TaskDefinition.name must match exactly. task_class_name is PascalCase+'Task'.
+3. **Instance completeness**: Every instance must have id and metadata dict.
```
georgosgeorgos, line 460: **[High] Shell quoting breaks for project paths containing spaces**

`_VALIDATE_TASK_COMMAND` embeds `{project_path}` inside an outer `bash -c '...'` single-quoted string. The executor substitutes via `shlex.quote(str(self.project_path))`, but `shlex.quote` on a path with spaces returns e.g. `'/Users/john smith/factory'` — this closes the outer single-quote prematurely, producing a shell syntax error.

This is common on macOS (`/Users/First Last/...`).

**Fix:** Use a shell variable instead of inline substitution so quoting never touches the `bash -c` string boundary:

```python
_VALIDATE_TASK_COMMAND = (
    "TASK_NAME=$(cat {project_path}/.factory/generated-task-name.txt) && "\n"factory task validate --name \"\\" --project {project_path}"
)
```

Or pass `project_path` as an env var:
```bash
PROJECT_PATH=/the/path bash -c 'factory task validate --name  --project ""'
```

.factory/workflows/task-setup.py
```diff
@@ -0,0 +1,1110 @@
...
+3. Generates improvement suggestions like "focus on reducing blunder_count"
+   or "improve coverage_pct" based on the differences it finds
+
+If your details dict is sparse (just `{passed: True, score: 0.8}`), the
+reflection mechanism has NOTHING to compare and evolution stalls. The details
+dict must contain rich, domain-specific numeric dimensions.
+
+### Design Guidance for VerifyResult.details
+
+- Include at least 4 numeric keys (int or float) — these enable quantitative
```
georgosgeorgos, line 513: **[Medium] GateNode.max_iterations is never read by the executor**

The executor enforces `verdict.max_iterations` (always 3, the `Verdict.reloop()` default) rather than `node.max_iterations`. Setting `max_iterations=2` on `gate_research` and `gate_qa` here has no effect — both gates will allow 3 reloops before halting.

This is a pre-existing executor bug (not introduced by this PR), but worth flagging since the intent here is clearly a 2-reloop cap. The tests assert the attribute value but not the enforced behavior, so this is silently wrong.

Until the executor is fixed to read `node.max_iterations`, the `max_iterations` declarations here are dead config. You may want to remove them to avoid misleading readers, or add a comment noting the current executor behavior.

.factory/workflows/task-setup.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    def __init__(self) -> None:
+        defn = TaskDefinition(
+            name="<task_name>",
+            description="<description>",
+            scoring=ScoringContract(method="<method>", metric_path="<path>"),
+            constraints=TaskConstraints(timeout=<seconds>, max_retries=<count>),
+        )
+        super().__init__(definition=defn)
+
```
georgosgeorgos, line 618: **[Low] Archivist `reads` set missing `.factory/generated-task-name.txt`**

The archivist prompt instructs the agent to read `.factory/generated-task-name.txt` to construct the archive entry, but the node's `reads` set doesn't declare it. `_wait_for_reads` won't gate on this file being available before the archivist starts.

In practice this is fine because `validate_task` runs synchronously before archivist. But if the workflow is later restructured (e.g. `validate_task` made non-blocking), the archivist could start before the file exists with no dependency tracking to catch it.

```python
reads={".factory/generated-task-name.txt", ".factory/strategy/qa-report.md", ...},
```

.factory/workflows/task-setup.py
```diff
@@ -0,0 +1,1110 @@
...
+    def instances(self) -> Iterator[TaskInstance]:
+        \"\"\"Discover what to work on.\"\"\"
+        yield TaskInstance(id="<id>", metadata={...})
```
georgosgeorgos, line 621: **[Low] Archivist writes to a fixed path, overwriting prior session data on re-runs**

The archivist always writes to `.factory/archive/task-setup.md`. Running the workflow twice silently overwrites the first session's domain findings, QA report, and audit trail.

Consider parameterising with the generated task name:
```python
writes={".factory/archive/task-setup-{task_name}.md"},
```

Or at minimum using a timestamp suffix so successive runs don't clobber each other.

georgosgeorgos (commented):
High-effort review of the task-setup workflow. Found 2 blocking issues and 4 lower-severity ones — all on the workflow file itself except one stale error message in `run.py`.

factory/cli/run.py
```diff
@@ -345,7 +345,7 @@ def cmd_run(args: argparse.Namespace) -> int:
             file=sys.stderr,
         )
         return 1
-    if focus and mode not in (*DESIGN_MODES, "research"):
+    if focus and mode not in (*DESIGN_MODES, "research", "task-setup"):
         print(
             f"Error: --focus (targeted mode) only works in design or research mode, got '{mode}'. "
```
georgosgeorgos: **[Low] Stale error message omits task-setup from valid focus modes**

This error message still lists only `design` and `research` as valid modes for `--focus`, but `task-setup` is also valid after this PR. Users debugging a focus-mode restriction will get inaccurate guidance.

One-liner fix: update the string to include `task-setup` in the list.

georgosgeorgos:
**[Low] Stale error message in `factory/cli/run.py:350` omits task-setup**

Not in this PR's diff, but related: the `--focus` mode validation error message lists only `design` and `research` as valid modes. After this PR, `task-setup` is also valid. A user who hits the error will get inaccurate guidance. One-liner fix in `run.py`.

georgosgeorgos:
@colehurwitz left a full review above — 2 blocking issues before this can merge (validate_task needs a RELOOP edge back to builder, and the bash quoting breaks on paths with spaces) plus 4 lower-severity ones. Happy to help fix any of these if useful.

.factory/workflows/task-setup.py
colehurwitz: Fixed in 4f6c15b6 — converted `validate_task` from FnNode to `GateNode(evaluator_type='fn')` with a RELOOP edge back to `builder` and PROCEED edge to `archivist`. Edge count 11 → 12.

.factory/workflows/task-setup.py
colehurwitz: Fixed in 4f6c15b6 — removed the `bash -c` wrapper. Now uses bare `{project_path}` placeholder that the executor handles via `shlex.quote()`.

.factory/workflows/task-setup.py
colehurwitz: Fixed in 4f6c15b6 — added advisory comments on the `max_iterations=2` lines noting that the executor currently uses `verdict.max_iterations` (default 3), not the node field.

.factory/workflows/task-setup.py
colehurwitz: Fixed in 4f6c15b6 — added `.factory/generated-task-name.txt` to the archivist's `reads` set.

.factory/workflows/task-setup.py
colehurwitz: Fixed in 4f6c15b6 — changed archivist `writes` from fixed path to directory (`.factory/archive/`) with dynamic filename based on task name.

factory/cli/run.py
colehurwitz: Fixed in 4f6c15b6 — updated error message in `run.py` to include `task-setup` in the valid modes list, plus updated the test assertion.

colehurwitz:
All 6 items addressed in 4f6c15b6 — replied to each inline comment above. The two blocking issues (validate_task RELOOP + shell quoting) and four lower-severity items are all fixed. Ready for re-review.

georgosgeorgos:
## Suggestion: split into two PRs

After reviewing this end-to-end, I'd suggest splitting it before merge. The current PR is doing two distinct things: (1) establishing the core scaffolding pipeline, and (2) adding a quality/validation layer on top of it. These have very different risk profiles and review concerns.

**PR A — Core scaffold (5 nodes):**
```
research_director → gate_research → strategy_director → gate_strategy (user) → builder → archivist
```
- Mandatory outer loop feedback researcher in research phase
- User approval gate after strategy
- Focused on getting the research → build pipeline right with minimal surface area
- 3 prompt templates to review carefully instead of 8

**PR B — Quality layer (adds 4 nodes on top of PR A):**
```
... → builder → qa_director → gate_qa → validate_task (GateNode, not FnNode) → archivist
```
- QA Director with 3 adversarial testers
- `validate_task` as `GateNode(evaluator_type="fn")` with a RELOOP edge back to `builder` — currently it's an FnNode with no recovery path, so any validation failure permanently halts the workflow
- Fix for the shell quoting issue in `_VALIDATE_TASK_COMMAND` (paths with spaces break `bash -c '...'`)

**Why split:**
The 8 prompt templates are the highest-risk part of this PR. They define complex multi-step agent behavior that's hard to evaluate by reading — you need to run it to know if the prompts produce good Task output. Splitting lets PR A land with 3 carefully reviewed prompts, get tested in real runs, and then PR B builds on validated behavior.

Also writing an ADR for the overall prompt design philosophy — will share separately.

colehurwitz:
> ## Suggestion: split into two PRs
> After reviewing this end-to-end, I'd suggest splitting it before merge. The current PR is doing two distinct things: (1) establishing the core scaffolding pipeline, and (2) adding a quality/validation layer on top of it. These have very different risk profiles and review concerns.
> 
> **PR A — Core scaffold (5 nodes):**
> 
> ```
> research_director → gate_research → strategy_director → gate_strategy (user) → builder → archivist
> ```
> 
> * Mandatory outer loop feedback researcher in research phase
> * User approval gate after strategy
> * Focused on getting the research → build pipeline right with minimal surface area
> * 3 prompt templates to review carefully instead of 8
> 
> **PR B — Quality layer (adds 4 nodes on top of PR A):**
> 
> ```
> ... → builder → qa_director → gate_qa → validate_task (GateNode, not FnNode) → archivist
> ```
> 
> * QA Director with 3 adversarial testers
> * `validate_task` as `GateNode(evaluator_type="fn")` with a RELOOP edge back to `builder` — currently it's an FnNode with no recovery path, so any validation failure permanently halts the workflow
> * Fix for the shell quoting issue in `_VALIDATE_TASK_COMMAND` (paths with spaces break `bash -c '...'`)
> 
> **Why split:** The 8 prompt templates are the highest-risk part of this PR. They define complex multi-step agent behavior that's hard to evaluate by reading — you need to run it to know if the prompts produce good Task output. Splitting lets PR A land with 3 carefully reviewed prompts, get tested in real runs, and then PR B builds on validated behavior.
> 
> Also writing an ADR for the overall prompt design philosophy — will share separately.

This is a single workflow to produce tasks which includes the planning, building, and QA. Not sure it totally makes sense to split into two PRs for a single workflow.

PR #1514, opened 2026-09-16
fix: add explicit synchronous execution instructions to director prompts (#1512)

### LINKED ISSUE #1512: bug: create-v2 CEO gets stuck polling for Strategy Director background task output
## Problem

Create-v2 and design-v2 modes occasionally get stuck when the CEO spawns director agents (Research Director, Strategy Director) or builder agents. The CEO enters a `while true` polling loop waiting for agent output files, and the polling condition sometimes never matches.

## What we know

- v1 create/design modes use `cmd1 & cmd2 & cmd3 & wait` in a single Bash call — works reliably
- v2 directors spawn sub-agents inside a nested Claude Code session — the `& wait` pattern behaves differently there
- The polling is the CEO's recovery strategy when it backgrounds an agent call
- A prompt-level fix (PR #1514, adding CRITICAL/SYNCHRONOUS blocks) was tried and reverted (PR #1518) — it was a symptom fix that didn't address the root cause and was Claude Code-specific
- The issue is NOT the Bash tool timeout — agents run in foreground for 9+ minutes without being auto-backgrounded

## Root cause (needs research)

Why does the CEO choose to background agent calls and poll in v2 but not v1? Both SKILL.md files use `& wait` for intentional parallelism. The difference may be in how nested agent sessions (directors spawning sub-agents) interact with shell backgrounding.

## Requirements for the fix

- Must not be Claude Code-specific (no references to Bash tool behavior)
- Must work for any orchestrator calling `factory agent`
- Must handle the case where agents take 5-15 minutes
- Should not break intentional parallelism (`& wait` pattern)

colehurwitz:
Closes #1512

## Changes

- Added CRITICAL synchronous execution instruction blocks to PHASE 2 sections of 4 director prompts across 2 files:
  - `CREATE_RESEARCH_DIRECTOR_PROMPT` in `factory/workflow/contributed/create_v2/prompts.py`
  - `CREATE_STRATEGY_DIRECTOR_PROMPT` in `factory/workflow/contributed/create_v2/prompts.py`
  - `RESEARCH_DIRECTOR_PROMPT` in `factory/workflow/contributed/design_v2/prompts.py`
  - `STRATEGY_DIRECTOR_PROMPT` in `factory/workflow/contributed/design_v2/prompts.py`
- Each block explicitly states that `factory agent` calls are synchronous (blocking), prohibits background execution with `&`, polling loops, and `grep -q` completion checks
- Includes FORBIDDEN anti-pattern example showing the exact hang pattern from #1512
- Includes CORRECT pattern showing the proper blocking call
- QA Director prompts intentionally unchanged — they correctly use `&` + `wait` for intentional parallelism
- Pure additions only — 84 lines added to prompt strings, no existing text modified
- All tests pass; ruff lint clean


colehurwitz:
@ceo-review

PR #1511, opened 2026-09-16
feat: add batch-summarizer contributed workflow

colehurwitz:
Closes #1508

## Changes

- Created `factory/workflow/contributed/batch_summarizer/` package with 5-node workflow graph:
  - **scan_dir** (FnNode) — scans source directory, generates JSONL manifest of document files
  - **data_loop** (DataNode) — iterates over manifest with `source_format='jsonl'`, runs subgraph per document (`parallelism=1`)
  - **summarize** (LLMNode) — reads document via `file_read`, writes structured Markdown summary via `file_write` (model: sonnet, max_turns: 5)
  - **write_summary** (FnNode) — fallback summary persistence if LLM's file_write call fails
  - **collect_index** (FnNode) — aggregates all per-file summaries into `SUMMARY_INDEX.md`
- 3 explicit edges: `scan_dir→data_loop`, `summarize→write_summary`, `data_loop→collect_index`
- Trigger: `mode=='batch-summarizer'`, `terminal=True`
- Registered in `_get_builtin_registry()` in `factory/workflow/definitions.py` via lazy import lambda
- `FILE_READ_TOOL` imported from `factory.workflow.llm_tools`; `FILE_WRITE_TOOL` defined inline as `ToolDef`
- `ProjectState` imported from `factory.workflow.primitives` (not `factory.models`)
- 29 structural tests across 7 test classes, all passing
- Graph validates: `factory workflow validate --file workflow.py` → VALID (5 nodes, 3 edges)
- Ruff lint: all checks passed
- README.md with usage documentation

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1510, opened 2026-09-16
feat: add batch-summarizer project-local workflow

colehurwitz:
Closes #1508

## Changes

- **`.factory/workflows/batch_summarizer.py`** — New project-local workflow with:
  - `DataNode` (`document_loader`): late-bound data source via `--data <path>` CLI flag. Uses `model_construct()` to bypass Pydantic's `_validate_source` check (requires exactly one source, but late-bound has zero)
  - `AgentNode` (`summarizer`): BUILDER role with prompt template for concise document summarization (3-5 sentences, key points, actionable insights)
  - `Workflow`: also constructed via `model_construct()` to prevent Pydantic's strict union discriminator from re-validating the late-bound DataNode
  - Zero edges — DataNode dispatches to its subgraph internally; single-node subgraph (entry == exit) needs no internal edges
  - `meta` dict with `name="batch-summarizer"` and description for registry discovery
- **`tests/test_batch_summarizer_workflow.py`** — 10 test cases: graph validation, start_node type, subgraph entry/exit, late-bound source fields, no DataNode-to-subgraph edges, meta dict, summarizer role, SKILL.md generation, name consistency, node count
- **SKILL.md** auto-generated via `factory workflow export-skills` (gitignored, not committed)

All 10 tests pass. Lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1508, opened 2026-09-15
feat: create mode --data-node and --task flags for DataNode/Task workflow creation

colehurwitz:
Closes #1505

## Changes

- **`--data-node` flag** (boolean, `dest="data_node"`): Added to the `ceo` subparser. Requires `--mode create`. Injects a focused directive telling the Builder to design a per-item subgraph pipeline. The Builder never sees the word "DataNode" — DataNode wrapping is mechanical post-processing via `wrap_with_data_node()`.

- **`--task TASK_REF` flag** (string, `dest="task_ref"`): Added to the `ceo` subparser. Requires `--mode create` but does NOT require `--data-node` (they are independent). Eagerly resolves the Task class via `_resolve_task_for_directive()` and injects field names, sample prompt, and scoring contract into the Builder directive.

- **Late-bound DataNode support**: Relaxed `DataNode._validate_source` from `sum(sources) != 1` to `sum(sources) > 1`, allowing zero sources for template/late-bound DataNodes.

- **`wrap_with_data_node()` utility** (`factory/workflow/wrapping.py`, NEW): Standalone mechanical wrapping that takes a Builder-produced subgraph workflow and wraps it with a DataNode. Handles terminal node detection, ID collision fallback, and optional `task_ref`.

- **Directive injection** uses additive `if`-blocks (not `elif`), supporting all 4 combinations: neither, `--data-node` only, `--task` only, both together. Zero context pollution when flags are absent.

### Files Changed

| File | Change |
|------|--------|
| `factory/cli/_parser_groups.py` | Two new `add_argument` calls |
| `factory/cli/ceo.py` | Flag extraction, validation, `_resolve_task_for_directive()` helper, threading |
| `factory/cli/_ceo_helpers.py` | Forward 3 new params in signature and call |
| `factory/cli/_task_builder.py` | 3 new params + two directive injection blocks |
| `factory/workflow/primitives.py` | Validator relaxation (`!= 1` → `> 1`) |
| `factory/workflow/wrapping.py` | **NEW** — `wrap_with_data_node()` |
| `tests/test_data_node.py` | Updated validation tests |
| `tests/test_wrapping.py` | **NEW** — 8 wrapping tests |
| `tests/test_cli.py` | 11 new CLI flag + directive injection tests |

### Test Results

All 359 tests pass (69 data_node + 282 CLI + 8 wrapping). Lint and type checks clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1494, opened 2026-09-10
fix: close execution-layer gaps for Task+Workflow consumers (#1488)

colehurwitz:
## Changes

- **Diagnostic logging in `_execute_data`** (`factory/workflow/executor.py`):
  - `data_item_setup_complete` info log after `setup()` — lists files created in workspace (up to 20)
  - `setup_read_path_mismatch` warning during setup_reads rescan — detects when a declared read path doesn't match the actual file location by searching for the basename recursively
  - `wait_for_reads_timeout_diagnostic` warning in `_wait_for_reads` — logs completed_files snapshot when timeout occurs, aiding debugging of stuck workflows

- **Integration test: DataNode + Loop(AgentNode body, fn GateNode)** (`tests/test_data_node.py`):
  - `test_data_node_loop_with_agent_body` — verifies 3-iteration RELOOP→PROCEED cycle with a mock agent_fn that appends to counter.txt
  - `test_setup_read_path_mismatch_logs_warning` — verifies that a mismatched read path causes inner executor halt (with fast timeout patch for CI)

All 747 workflow tests pass (69 in test_data_node.py).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

**Original scope:** Close 3 execution-layer gaps from #1488 so consumers can use `(Task, Workflow) → SwarmEngine` without adapter code.

**Fixes included:**
- compose() skips capability validation for DataNode workflows
- WorkflowExecutor.\_run_agent() persists output to node.writes
- WorkflowExecutor accepts agent_fn parameter (propagates to sub-executors)
- Re-scan disk reads after task.setup() in DataNode executor
- NoveltyFilter skips edit-distance when structural hash differs
- Designer preserves frozen nodes in variants
- Diagnostic logging + loop integration test

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## All fixes included in this PR

1. **compose() skips capability validation for DataNode workflows** — DataNode workflows are evaluation-only, don't need HAS_BUILDER/CAN_RUN_TESTS
2. **WorkflowExecutor._run_agent() persists output to node.writes** — fixes silent score=0 when Task.verify() reads from disk
3. **WorkflowExecutor accepts agent_fn parameter** — propagates to SubgraphFork and DataNode sub-executors
4. **Re-scan disk_reads after task.setup()** — setup-created files now visible to sub-executor's _wait_for_reads
5. **NoveltyFilter skips edit-distance when structural hash differs** — prompt mutations no longer rejected as near-duplicates
[...]

PR #1515, opened 2026-09-16
fix: untrack .factory/reviews/builder-latest.md

colehurwitz:
## Problem

`.factory/reviews/builder-latest.md` was accidentally committed to main in PR #1494. Once a file is tracked, `.gitignore` no longer applies — so every PR whose CEO cycle writes to this file creates a merge conflict.

This has been causing recurring conflicts on PRs #1507, #1508, #1510, #1511, #1514, etc.

## Fix

`git rm --cached` removes the file from the index without deleting it from disk. After this merges, `.gitignore` takes effect again and the file will never be committed.

## Root cause

PR #1494's Builder committed `.factory/reviews/builder-latest.md` alongside the source changes. The `.gitignore` rule for `.factory/` was already present but has no effect on tracked files.


colehurwitz:
@ceo-review

PR #1518, opened 2026-09-16
revert: remove symptom fix for director polling (#1514)

colehurwitz:
## Why

PR #1514 added CRITICAL/FORBIDDEN blocks to director prompts telling the CEO not to background agent calls. This was a **symptom fix** — the real problem is that `factory agent` calls take longer than the Bash tool timeout (120s), causing the orchestrator to background them and then poll for completion.

The prompt-level fix:
- Only covered directors, not Builder or other agent spawns
- Is Claude Code-specific (references Bash tool behavior)
- Doesn't prevent the timeout from firing — the orchestrator will still background long calls regardless of prompt instructions

The real fix needs to be at the factory CLI level — a proper completion signal mechanism that any orchestrator can use, not just Claude Code.

See #1512 for the updated root cause analysis.


colehurwitz:
@ceo-review

PR #1519, opened 2026-09-16
fix: add backgrounding recovery instructions to CEO prompt

### LINKED ISSUE #1512: bug: create-v2 CEO gets stuck polling for Strategy Director background task output

colehurwitz:
Closes #1512

## Changes

- **Updated polling instruction (line 79):** Replaced blanket "do not poll for subagent completion" with specific guidance — do not poll using Claude Code's internal task output file; instead, poll for the node's output artifacts when auto-backgrounding occurs
- **Added Backgrounding Recovery subsection:** Documents the correct recovery pattern when Claude Code auto-backgrounds a command after 600s timeout:
  - Poll for the output artifact listed in the SKILL.md verification block (file existence + minimum size check)
  - Explicitly forbids grepping Claude Code's internal task output file (fragile string matching causes infinite loops, which is the root cause of #1512)
  - Instructs the CEO to proceed to artifact verification after the artifact appears

Only `factory/agents/prompts/ceo.md` is modified. No changes to director prompts, workflow files, or any other files.


colehurwitz:
## ✅ KEEP — Factory CEO Approval

**Experiment:** Fix bug #1512 — CEO polling recovery after auto-backgrounding
**Verdict:** KEEP
**PR:** #1519

### Evidence

| Signal | Result |
|--------|--------|
| Tests | ✅ 165 passed |
| Lint | ✅ Clean |
| Type check | ✅ Clean |
| Composite score | 0.5915 (identical to baseline) |
| Score delta | 0.0000 (prompt-only change) |
| Guard violations | None |
| Code review | CLEAN (7/7 pass) |
| Adversarial QA | PASS (8/8 criteria verified) |

### Root Cause (from investigation)
Claude Code's Bash tool auto-backgrounds commands exceeding 600s. When `factory agent ceo --task "Director..."` is backgrounded, the CEO must poll for completion. The bug was that the CEO sometimes used a **fragile grep on Claude Code's internal task output file** instead of polling for the **known output artifact** the node writes.

### Fix
- Updated line 79 in `ceo.md`: replaced blanket "Do NOT poll" with specific guidance to poll for output artifacts
- Added "Backgrounding Recovery" subsection: shows correct artifact-polling pattern + forbids task-output-file grep

### Decision Rationale
- **Minimal scope:** 1 file, 22 insertions, 1 deletion
- **Evidence-based:** compared successful vs failing polling patterns from 5+ actual session transcripts
- **Positive instruction:** shows exact correct pattern (what PR #1514 tried with negative prohibitions and failed)
- **No regression:** zero eval score impact (prompt-only change)

Closes #1512

colehurwitz:
@ceo-review

PR #1522, opened 2026-09-16
fix: bidirectional BFS in _collect_subgraph_nodes to exclude stray branches

### LINKED ISSUE #1246: bug: _collect_subgraph_nodes includes nodes not on any entry→exit path
## Current Behavior

`_collect_subgraph_nodes(entry, exit_node)` is documented as "Collect all node IDs on paths from entry to exit_node (inclusive)" but implements forward BFS only: it starts at entry, visits every reachable successor, and stops *expanding* only when node == exit. Nodes reachable from entry but NOT on any path to exit (e.g. a stray branch) are still collected and returned.

## Desired Behavior

Only nodes on at least one entry→exit path should be returned: `reachable_from_entry ∩ can_reach_exit` (plus entry/exit themselves). Nodes that cannot reach exit should be excluded, so the extracted sub-workflow matches the documented semantics.

## Evidence

`factory/workflow/executor.py:1111-1135` (current main):

```python
def _collect_subgraph_nodes(
    workflow: Workflow,
    entry: str,
    exit_node: str,
) -> set[str]:
    """Collect all node IDs on paths from entry to exit_node (inclusive)."""
    edges_by_source: dict[str, list[str]] = {}
    for edge in workflow.edges:
        edges_by_source.setdefault(edge.source, []).append(edge.target)

    # BFS from entry, stop at exit_node
    visited: set[str] = set()
    queue = [entry]
    while queue:
        nid = queue.pop(0)
        if nid in visited:
            continue
        visited.add(nid)
        if nid == exit_node:
            continue
        for target in edges_by_source.get(nid, []):
            if target not in visited:
                queue.append(target)

    return visited
```

Repro: workflow with edges `entry → a → exit` and `entry → stray → x`. Calling `_collect_subgraph_nodes(wf, "entry", "exit")` returns `{'a', 'entry', 'exit', 'stray', 'x'}` — `stray` and `x` are on no path to exit, yet are included.

Impact: the collected set is passed to `Workflow.subgraph()` (`factory/workflow/primitives.py:287-309`), which deep-copies all nodes in the set into the branch workflow — so stray nodes get executed in every parallel experiment branch even though they are not part of the entry→exit subgraph.

## Call Trace

1. `WorkflowExecutor.execute()` → `_execute_from(start_node)` (`factory/workflow/executor.py:181`)
2. `_execute_from` dispatches `SubgraphForkNode` → `await self._execute_subgraph_fork(node)` (`factory/workflow/executor.py:196-198`)
3. `_execute_subgraph_fork` (`factory/workflow/executor.py:487`) calls `_collect_subgraph_nodes(self.workflow, node.subgraph_entry, node.subgraph_exit)` (`factory/workflow/executor.py:533-535`)
4. `_collect_subgraph_nodes` (`factory/workflow/executor.py:1111-1135`) — buggy forward-only BFS
5. Result feeds `self.workflow.subgraph(subgraph_ids, ...)` (`factory/workflow/executor.py:536`); `Workflow.subgraph` (`factory/workflow/primitives.py:287-309`) deep-copies every collected node and filters edges to the set, so stray-branch nodes become part of the sub-workflow that each branch executor runs
6. Second caller with same bug: `factory/workflow/skill_export.py:815-816` — stray nodes get treated as subgraph nodes and excluded from the linear skill phases

georgosgeorgos:
Closes #1246

## Problem

`_collect_subgraph_nodes(entry, exit_node)` used a forward-only BFS from `entry`, collecting every reachable node and stopping *expansion* (but not *collection*) at `exit_node`. Nodes on stray branches — reachable from `entry` but not leading to `exit_node` — were incorrectly included.

## Fix

Bidirectional BFS: intersect nodes reachable **forward from entry** with nodes reachable **backward from exit_node**. Only nodes on at least one `entry → exit_node` path are returned.

## Tests

4 new tests in `TestCollectSubgraphNodes`: linear chain, stray branch excluded, parallel branches both included, entry-equals-exit edge case.


georgosgeorgos:
@colehurwitz would love your review on this one when you get a chance.

georgosgeorgos:
@lambdabaa would love your review on this too.

colehurwitz:
# PR Review: #1522 - fix: bidirectional BFS in _collect_subgraph_nodes to exclude stray branches

Reviewed: 2026-09-17

## Summary
PR #1522 fixes a bug in `_collect_subgraph_nodes()` where a forward-only BFS collected all nodes reachable from `entry`, including nodes on stray branches that don't lead to `exit_node`. The fix replaces the single forward BFS with a bidirectional BFS approach: intersect nodes reachable forward from `entry` with nodes reachable backward from `exit_node`. This ensures only nodes on at least one `entry → exit_node` path are returned. The duplicated implementation in `validation.py` is replaced with a delegation to the executor's fixed version. Four new unit tests cover linear chains, stray branch exclusion, parallel convergent branches, and the `entry == exit` edge case.

## Root Cause Analysis
- **Finding:** This PR addresses the **root cause** — an algorithmic mismatch between the documented semantics ("collect all node IDs on paths from entry to exit_node") and the implementation (forward-only BFS that collects all nodes reachable from entry, regardless of whether they reach exit_node).
- **Evidence:**
  - The old code at `executor.py:1576-1590` uses forward-only BFS. The line `if nid == exit_node: continue` stops *expansion* beyond exit_node but does NOT filter out nodes that cannot reach exit_node. This is the fundamental algorithmic error.
  - The fix at lines 1569-1601 (in the diff) implements the mathematically correct solution: `forward_reachable(entry) ∩ backward_reachable(exit_node)`. This is the standard graph-theoretic approach for finding nodes on paths between two vertices.
  - This is NOT a symptom-level fix. It doesn't add a try/except, a null-check, a retry, or a special case. It replaces the incorrect algorithm with the correct one.
  - The fix also removes code duplication by having `validation.py` delegate to the single corrected implementation in `executor.py`.
- **Rating:** ✅ Addresses root cause

## Solution Validation
- **Finding:** The PR **solves the stated problem** completely for the three call sites that use `_collect_subgraph_nodes`.
- **Evidence:**
  1. **Issue requirement: "Only nodes on at least one entry→exit path should be returned: `reachable_from_entry ∩ can_reach_exit`"** — The bidirectional BFS intersection (`forward & backward`) at `executor.py` line 1601 (diff) implements this exact formula. ✅
  2. **Issue call site 1 — `_execute_subgraph_fork` (`executor.py:567`)** — Directly calls `_collect_subgraph_nodes` from `executor.py`, which is fixed. ✅
  3. **Issue call site 2 — `_execute_data` (`executor.py:785`)** — Same direct call, fixed. ✅
  4. **Issue call site 3 — `skill_export.py:731,739`** — Already imports `_collect_subgraph_nodes` from `factory.workflow.executor`, so it inherits the fix automatically. ✅
  5. **Issue call site 4 — `validation.py:143`** — The duplicated buggy implementation in `validation.py:185-208` is replaced with delegation to the executor's fixed version. ✅
  6. **Repro scenario from issue** — "workflow with edges `entry → a → exit` and `entry → stray → x`" — This is directly tested in `test_stray_branch_excluded` which asserts `stray` is not in the result. ✅
  7. **Tests verify correctness** — 4 tests cover: linear chain (basic path), stray branch exclusion (the core bug), parallel branches (both converging paths included), and entry==exit edge case. ✅

- **Edge cases examined:**
  - **`entry == exit`**: Tested explicitly in `test_entry_equals_exit`. Both forward BFS from `a` and backward BFS from `a` return `{a}`, intersection is `{a}`. Correct. ✅
  - **Disconnected exit_node (exit not reachable from entry)**: Not tested. Forward BFS won't include exit_node, backward BFS won't include entry. Intersection would be empty — `entry` and `exit_node` are both excluded. This could be problematic if callers assume entry/exit are always in the result, but the validation step (`validate_workflow`) ensures paths exist, so this case is guarded upstream. ⚠️ Minor gap.
  - **Cycles in the graph**: The BFS `visited` set prevents infinite loops. The `stop_at` parameter prevents expanding beyond the boundary node. No issue. ✅
  - **Large graphs / performance**: Replacing `list.pop(0)` (O(n)) with `deque.popleft()` (O(1)) is a performance improvement. Two BFS passes is still O(V+E). ✅

- **Potential issues introduced:**
  - **Circular import risk**: `validation.py` now does `from factory.workflow.executor import _collect_subgraph_nodes`. This is a lazy import inside a function body (line 190 of the diff), not a top-level import, so circular import is avoided. The existing `skill_export.py` already uses the same pattern (lazy import from executor inside function body at lines 731, 737). ✅ No issue.
  - **Private function exported across modules**: `_collect_subgraph_nodes` is prefixed with `_` (private convention) but is imported by `validation.py` and `skill_export.py`. This is an existing pattern, not introduced by this PR. ⚠️ Minor style concern, pre-existing.

- **Rating:** ✅ Solves problem

## Recommendations

1. **Consider adding a test for disconnected entry/exit** (minor, non-blocking): A test where `exit_node` is not reachable from `entry` would document the behavior (empty set returned). While validation guards this upstream, the function's contract should be clear about this edge case.
   - File: `tests/test_workflow_executor.py`, after `test_entry_equals_exit`

2. **Consider testing the validation.py delegation** (minor, non-blocking): While existing validation tests provide implicit coverage, a direct test verifying `validation._collect_subgraph_nodes` returns the same result as `executor._collect_subgraph_nodes` would guard against future divergence.

3. **File length check**: `executor.py` is at 1590 lines with the old code. The PR adds 11 net lines (28 additions - 17 deletions), bringing it to ~1601 lines. This exceeds the 500-line guardrail significantly, but this is a pre-existing condition not introduced by this PR. The PR's net change is minimal (+11 lines).

## Approval Status
✅ Approve

The PR correctly identifies and fixes the root cause (forward-only BFS algorithm), implements the mathematically correct solution (bidirectional BFS intersection), eliminates code duplication (validation.py delegates to executor.py), and includes thorough regression tests. All three call sites (`_execute_subgraph_fork`, `_execute_data`, and `skill_export`) benefit from the fix. The algorithm change is clean and well-documented in the updated docstring. No new risks are introduced.

colehurwitz:
@ceo-review

georgosgeorgos:
@colehurwitz all CI green, local tests pass across 3.11/3.12/3.13. Branches are now clean and independent — can be merged in any order. Ready for approval.

PR #1524, opened 2026-09-16
fix: JoinNode validates declared sources before proceeding

### LINKED ISSUE #1189: bug: JoinNode never waits for its sources — barrier semantics unimplemented
## Current Behavior

`JoinNode` is documented as a barrier — wait for all parallel branches to complete before proceeding — and carries a `sources` field for exactly that purpose. But the deterministic executor never checks it. The handler immediately marks the node complete and continues:

```python
if isinstance(node, JoinNode):
    self.result.nodes_executed += 1
    self.completed_files |= node.writes
    next_id = self._next_unconditional(node_id)
    if next_id:
        await self._execute_from(next_id)
    return
```
(factory/workflow/executor.py:207-213)

`node.sources` is never read at runtime. A join wired to sources that never ran would pass through instantly, as long as it has an outgoing edge. Barrier semantics currently work only by accident: the preceding `ForkNode` awaits its branches before control flow reaches the join (executor.py:415-484), so the join never gets a chance to fail.

## Desired Behavior

`JoinNode` should machine-enforce waiting: before proceeding, verify that all declared `sources` have completed (and ideally that their declared writes are satisfied). The barrier should be real, not inherited from the fork's behavior.

georgosgeorgos:
Closes #1189

## Problem

`JoinNode` in the deterministic executor never consulted `node.sources`. It immediately incremented `nodes_executed`, updated `completed_files`, and followed the outgoing edge — making JoinNode a no-op barrier. Any execution-ordering violation (JoinNode reached before parallel branches finished) was silently ignored.

## Fix

Before proceeding, JoinNode checks that every node ID in `node.sources` is present in `self.result.node_outputs`. If any are missing:
- Workflow halts with `halt_reason` naming the pending sources
- Structured log event `joinnode.sources_incomplete` emitted with pending/completed lists

Empty `sources` lists still proceed unconditionally (backward compat for JoinNodes that don't declare sources).

## Tests

3 new tests in `TestJoinNodeBarrier`:
- All sources complete → succeeds
- Missing source → halts with descriptive message naming the missing node
- Empty sources → proceeds unconditionally

Existing `TestForkJoin::test_fork_runs_concurrently` still passes — after `ForkNode.asyncio.gather()` all branch outputs are in `node_outputs` before JoinNode is reached.


georgosgeorgos:
@colehurwitz would love your review on this one when you get a chance.

georgosgeorgos:
@lambdabaa would love your review on this too.

colehurwitz:
Reviewed: 2026-09-17

# PR Review: #1524 - fix: JoinNode validates declared sources before proceeding

## Summary
This PR closes issue #1189 by making `JoinNode` actively enforce its `sources` contract at runtime. Previously, `JoinNode` was a no-op pass-through in the deterministic executor — it never inspected `node.sources` and immediately proceeded. Barrier semantics only worked by accident because the upstream `ForkNode.asyncio.gather()` completed all branches before control flow reached the join. This PR adds source-completion checking, halts with a descriptive error on violations, and adds complementary fixes (GateNode/JoinNode self-registration in `node_outputs`, code deduplication in `validation.py`). Six new tests cover the core barrier semantics.

## Root Cause Analysis
- **Finding:** ✅ This PR addresses the root cause directly.
- **Evidence:**
  - The root cause was an **architectural gap**: `JoinNode`'s runtime handler (executor.py:231-237 old) treated the node as a marker, not a barrier. It incremented `nodes_executed`, updated `completed_files`, and followed outgoing edges — never reading `node.sources`. The barrier property was accidentally inherited from `ForkNode.asyncio.gather()`, creating a control-flow dependency inversion.
  - The fix adds an explicit source-completion check (executor.py:232-261 new) that validates every declared source is present in `self.result.node_outputs` before proceeding. This makes the barrier an **intrinsic property of JoinNode** rather than a side effect of upstream execution.
  - The fix is not a symptom-level workaround — it doesn't add retries, timeouts, or error suppression. It implements the contract that `node.sources` was always intended to enforce, as documented in the `JoinNode` docstring ("Barrier node — waits for all sources to complete") and the validation layer (validation.py:111-113 checks sources exist as nodes).
  - Complementary changes (GateNode and JoinNode self-registration in `node_outputs`) prevent false-halts that would arise from the barrier check. These are not symptoms fixes — they're completing the `node_outputs` tracking contract for node types that previously skipped it.
- **Rating:** ✅ Addresses root cause

## Solution Validation
- **Finding:** ✅ This PR solves the stated problem completely.
- **Evidence:**

  **Issue requirement 1: "JoinNode should machine-enforce waiting — verify that all declared sources have completed"**
  - ✓ Implemented: `pending = [s for s in node.sources if s not in self.result.node_outputs and getattr(self.workflow.nodes.get(s), "blocking", True)]` (executor.py:234-238)
  - Sources are checked against `node_outputs`, which is the definitive completion registry
  - Non-blocking (background) sources are correctly exempted via the `blocking` attribute check

  **Issue requirement 2: "The barrier should be real, not inherited from the fork's behavior"**
  - ✓ The check is in the JoinNode handler itself, independent of ForkNode. A JoinNode reached via any path (not just after a fork) will enforce its sources.

  **Issue requirement 3: "(ideally) verify that their declared writes are satisfied"**
  - ⚠️ The "ideally" parenthetical is not implemented. The check verifies source *completion* (presence in `node_outputs`) but not that the sources' declared writes are in `completed_files`. However, this was explicitly marked as optional ("ideally"), and the existing `_wait_for_reads()` mechanism (executor.py:215) already handles file-based dependencies for all nodes including JoinNode. This is a reasonable design choice, not a gap.

  **Backward compatibility:**
  - ✓ Empty `sources` lists proceed unconditionally (tested by `test_join_with_no_sources_declared_proceeds`)
  - ✓ Existing fork-join patterns work because `asyncio.gather()` ensures all fork targets are in `node_outputs` before the join is reached

  **Edge cases handled:**
  - ✓ Non-blocking sources exempted from barrier check (via `blocking` attribute)
  - ✓ GateNode as source works (GateNode now self-registers in `node_outputs`)
  - ✓ JoinNode self-registers in `node_outputs` (prevents false-halt for nested joins or joins listing a join as source)
  - ✓ `node.failed` event emitted on barrier violation (consistent with other halt paths)

  **Remaining gaps (minor, non-blocking):**
  - No test for non-blocking sources exemption (the code handles it, but no test exercises this path)
  - `ForkNode` and `SelectionNode` still don't register themselves in `node_outputs`. If a JoinNode listed a ForkNode in `sources`, it would false-halt. This is an unlikely topology (joins list fork *targets* as sources, not the fork itself), but the `node_outputs` contract is still inconsistent across node types.
  - The `_collect_subgraph_nodes` deduplication introduces a module-level dependency direction from validation.py → executor.py. The local import avoids circular dependency, but it's slightly unusual. This is a style choice, not a correctness issue.

- **Rating:** ✅ Solves problem

## Recommendations

1. **Minor (non-blocking):** Add a test for the non-blocking source exemption path. The `getattr(self.workflow.nodes.get(s), "blocking", True)` logic at executor.py:238 is untested. A test with a non-blocking source in `JoinNode.sources` would verify it doesn't false-halt.

2. **Minor (non-blocking):** Consider registering `ForkNode` in `node_outputs` (even with an empty string) for consistency with the new JoinNode and GateNode registrations. This would prevent hypothetical false-halts if a join ever lists a fork as a source, and make the `node_outputs` contract uniform across all node types. This can be done in a follow-up PR.

3. **Nit:** The commit message for commit 2/3 mentions "use deque in _bfs" but the actual code on the PR branch (executor.py:1613) still uses `queue.pop(0)` rather than `collections.deque` + `popleft()`. The commit message is slightly misleading, though the functional behavior is correct.

## Approval Status
✅ **Approve**

The PR correctly identifies and fixes the root cause (JoinNode handler never checking `node.sources`), implements the fix cleanly with proper error handling and observability (structured logs + events), maintains backward compatibility, and includes comprehensive tests covering happy path, failure path, backward compat, and edge cases. The complementary changes (GateNode/JoinNode self-registration) are necessary consequences of the barrier check, not scope creep. Minor gaps (non-blocking source test, ForkNode registration) are not blocking.

colehurwitz:
@ceo-review

georgosgeorgos:
@colehurwitz all CI green, local tests pass across 3.11/3.12/3.13. Branches are now clean and independent — can be merged in any order. Ready for approval.

colehurwitz (approved):

PR #1529, opened 2026-09-17
fix: DesignerAgent creates broken workflow variants (missing prompt_template, missing gate edges)

### LINKED ISSUE #1526: bug: DesignerAgent creates broken workflow variants (missing prompt_template, missing gate edges)
## Problem

The outer loop's `DesignerAgent` creates workflow variants that pass `validate_graph()` but crash at runtime. Two specific failures:

### 1. AgentNode with no prompt_template → agent crashes

`AgentNode(role=BUILDER)` with no `prompt_template`. The executor invokes `claude --print` with no prompt → `Error: Input must be provided either through stdin or as a prompt argument when using --print` → exit code 1.

### 2. GateNode with no PROCEED edge → workflow dead-ends

`GateNode(id="gate_qa")` with no PROCEED edge. The workflow can't advance past the gate.

## Evidence

From a doc-quality-scorer outer loop run:
- 5 `builder --print` failures (all in Designer's `minimal_docquality` variant)
- 4 `gate_proceed_edge_missing` (all in Designer's `minimal_docquality` variant)
- Seed workflow had **zero errors**
- These waste population slots and evaluation budget per generation

## Root cause

`validate_graph()` in `factory/workflow/validation.py` checks structural correctness (reachability, edges, cycles, fork/join, DataNode wiring) but NOT semantic completeness:

| Checked | Not checked |
|---|---|
| start_node exists | AgentNode has prompt_template |
| edges reference valid nodes | GateNode has PROCEED edge |
| all nodes reachable | AgentNode has reads or prompt (something to work with) |
| no invalid cycles | |
| fork/join match | |
| DataNode edges valid | |

The graph is structurally valid but not runnable.

### 1. Add semantic validation to `validate_workflow()` in `validation.py`

Two new checks:

```python
def _validate_agent_prompts(workflow: Workflow, issues: list[str]) -> None:
    """Every AgentNode must have a prompt_template or reads (something to work with)."""
    for nid, node in workflow.nodes.items():
        if type(node).__name__ == "AgentNode":
            if not getattr(node, "prompt_template", None) and not node.reads:
                issues.append(
                    f"AgentNode '{nid}' has no prompt_template and no reads — "
                    f"the agent will have nothing to work with at runtime"
                )

def _validate_gate_edges(workflow: Workflow, issues: list[str]) -> None:
    """Every GateNode must have at least one PROCEED edge."""
    gate_ids = {nid for nid, node in workflow.nodes.items() 
                if type(node).__name__ == "GateNode"}
    for gate_id in gate_ids:
        has_proceed = any(
            e.source == gate_id and e.condition and e.condition.value == "proceed"
            for e in workflow.edges
        )
        if not has_proceed:
            issues.append(
                f"GateNode '{gate_id}' has no PROCEED edge — "
                f"the workflow will dead-end at this gate"
            )
```

### 2. Outer loop uses validation before evaluation

In `SwarmEvaluator`, before spending agent budget on a candidate:

```python
issues = candidate_workflow.validate_graph()
if issues:
    return EvalResult(benchmark_score=0.0, details={"validation_errors": issues})
```

This is the same pattern as `frozen_node_violated` — fast-reject broken candidates without running them.

### 3. DesignerAgent creates valid variants

After `_inject_frozen_nodes()` and `_rewire_data_nodes()`, call `validate_graph()` on the result. If validation fails, attempt to fix (add default prompt from seed workflow, wire missing gate edges) or discard.

## Impact

- Eliminates wasted budget on guaranteed-to-fail variants
- Catches broken workflows at creation time, not at runtime
- Same validation used by create mode (`factory workflow validate`) and outer loop

colehurwitz:
## fix: DesignerAgent creates broken workflow variants (missing prompt_template, missing gate edges)

Closes #1526

### Root Cause

The `DesignerAgent` creates workflow variants optimized for the CEO+SKILL.md execution path, where `prompt_template` is unused — the CEO reads the SKILL.md playbook and crafts agent tasks itself. But when the outer loop evaluates these variants via `WorkflowExecutor` (the Task-based / headless path), the executor reads `node.prompt_template` directly and passes it to `claude --print`. An empty `prompt_template` means an empty prompt, which causes `claude --print` to exit with code 1: *"Input must be provided either through stdin or as a prompt argument when using --print."*

The second failure mode is subtler. The Designer templates correctly create `gate_qa` as a **terminal node** (zero outgoing edges) — this is valid for the SKILL.md path where the gate is the last step. But `_rewire_data_nodes()` converts `gate_qa` from a terminal node into a `subgraph_exit` inside a `DataNode`, changing its execution context. Inside a DataNode subgraph, the executor expects a `PROCEED` edge to advance past the gate after evaluation — but `_rewire_data_nodes()` never adds one. The workflow dead-ends.

### Evidence

From the `doc-quality-scorer` outer loop run (the first project to use `WorkflowExecutor`-based evaluation with the base `DesignerAgent`):

- **5 `builder --print` failures** — all in the Designer's `minimal_docquality` variant, all caused by empty `prompt_template` on AgentNodes
- **4 `gate_proceed_edge_missing`** — all in the same variant, caused by `gate_qa` having no `PROCEED` edge after `_rewire_data_nodes()` converted it to a subgraph exit
- **Seed workflow: 0 errors** — the seed had correct prompts and was not rewired

### Why This Was Never Seen Before

- **FeatureBench** uses the CEO+SKILL.md path exclusively — `prompt_template` is never read by the executor, so empty prompts are harmless
- **chess-evolve** uses `WorkflowExecutor` but has a **custom `ChessDesignerAgent`** that manually patches prompts onto AgentNodes after template creation — it accidentally worked around the base Designer's bug
- **doc-quality-scorer** was the first project to use the base `DesignerAgent` with `WorkflowExecutor`-based evaluation, exposing both failure modes

### Fix: 5-Part Layered Defense

Two **root fixes** that prevent the bugs, plus three **safety nets** that catch edge cases:

#### Root Fixes (prevent the problem)

**Part C — Prompt Propagation** (`designer.py`): New `_propagate_prompts_from_seed(nodes, seed_workflow)` helper. Builds a `{AgentRole → prompt_template}` lookup from the seed workflow and copies prompts to template AgentNodes by matching role. Called before `_inject_frozen_nodes` in all three design methods (`design_minimal`, `design_thorough`, `design_custom`), so frozen nodes always take priority over propagated prompts. No-op when `seed_workflow is None`.

**Part D — Gate Edge Fix** (`designer.py`): In `_rewire_data_nodes()`, after setting `subgraph_exit` to a terminal node, checks if that node is a `GateNode` with no `PROCEED` edge. If so, adds `Edge(source=gate_id, target=next_after_data, condition=VerdictType.PROCEED)`. The Designer templates are **not modified** — `gate_qa` as a terminal node remains correct for the SKILL.md path. The fix targets the rewiring logic that changes the gate's context.

#### Safety Nets (catch edge cases)

**Part A — Semantic Validation** (`validation.py`): Two new validators registered in `validate_workflow()`:
- `_validate_agent_prompts`: flags AgentNodes with empty `prompt_template`
- `_validate_gate_edges`: flags non-terminal GateNodes (those with outgoing edges) that lack a `PROCEED` or unconditional edge. Terminal gates (zero outgoing edges) are explicitly skipped — they are valid end-of-workflow sentinels.

**Part B — Evaluator Pre-check** (`evaluator.py`): New CHECK 4 in `SwarmEvaluator.evaluate()`, after the existing frozen-node check and before inner-loop dispatch. Calls `workflow.validate_graph()` and fast-rejects broken workflows with `EvalResult(score=0.0, details={"rejected": "graph_validation_failed"})` instead of burning a full inner-loop evaluation.

**Part E — `_validate_and_fix` Fallback** (`designer.py`): Last-resort safety net called after `Workflow(...)` construction in all three design methods. Fills empty `prompt_template` fields with a generic default and adds missing `PROCEED` edges to non-terminal gates. Catches cases the root fixes miss (e.g., novel agent roles not present in the seed workflow).

### Call Order in Each Design Method

```
1. Build template nodes + edges                  (existing)
2. _propagate_prompts_from_seed(nodes, seed)     ← Part C: ROOT FIX
3. _inject_frozen_nodes(nodes, edges, seed, ...) (existing)
4. _rewire_data_nodes(nodes, edges, ...)         ← Part D: ROOT FIX (inside this fn)
5. wf = Workflow(...)                            (existing)
6. wf = _validate_and_fix(wf, seed)              ← Part E: SAFETY NET
7. return wf
```

### Test Coverage

**32 new tests** across 6 test classes:

| Class | Tests | Covers |
|-------|-------|--------|
| `TestValidateAgentPrompts` | 5 | Empty prompt flagged, non-empty OK, non-agent skipped, mixed, multiple |
| `TestValidateGateEdges` | 6 | PROCEED OK, unconditional OK, RELOOP-only flagged, terminal skipped, both OK, HALT-only flagged |
| `TestPromptPropagation` | 9 | Fill empty, no-op None, unmatched role, same role ×2, existing kept, all 3 design methods, frozen wins |
| `TestGateEdgeWiring` | 3 | PROCEED added after rewire, non-gate no-op, idempotent |
| `TestValidateAndFixFallback` | 5 | No-op valid, fill prompt, add PROCEED, preserve existing, idempotent |
| `TestGraphValidationPreCheck` | 3 | Invalid rejected, valid passes, eval fn never called for invalid |

**Results:** 127/127 target tests pass, 621/621 full suite, mypy clean, ruff clean, composite score 0.9633 (zero regression).

### Files Changed

| File | Changes |
|------|---------|
| `factory/workflow/validation.py` | +35 — `_validate_agent_prompts`, `_validate_gate_edges`, registered in `validate_workflow()` |
| `factory/outer_loop/designer.py` | +144 — `_propagate_prompts_from_seed`, gate edge fix in `_rewire_data_nodes`, `_validate_and_fix` |
| `factory/outer_loop/evaluator.py` | +16 — CHECK 4 graph validation pre-check |
| `factory/workflow/definitions.py` | +19 — default prompts for health_checker, code_reviewer, adversarial_tester |
| `tests/test_outer_loop/conftest.py` | +4 — prompt_template + PROCEED edge in simple_workflow fixture |
| `tests/test_outer_loop/test_designer.py` | +315 — TestPromptPropagation, TestGateEdgeWiring, TestValidateAndFixFallback |
| `tests/test_outer_loop/test_evaluator.py` | +63 — TestGraphValidationPreCheck, fixture update |
| `tests/test_workflow_primitives.py` | +213 — TestValidateAgentPrompts, TestValidateGateEdges |


colehurwitz:
Superseded by #1533. Validation will be part of the configurable execution strategy PR.

